PR #1, opened 2023-07-24
Parker Boeing

ParkerBoeing:
I made it through all stories and extensions but I'm still unsure of active record versus ruby so fingers crossed the two words I put in the model methods are active record and not ruby. Pass, fail, who knows, you keep us on our toes!! 🤷‍♂️

### DIFF
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,0 +1,6 @@
+class GuestRoomsController < ApplicationController
+  def create
+    @guest_room = GuestRoom.create!(guest_id: params[:id], room_id: params[:room_id])
+    redirect_to "/guests/#{@guest_room.guest_id}"
+  end
+end
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,0 +1,6 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
+  end
+end
--- a/app/controllers/hotels_controller.rb
+++ b/app/controllers/hotels_controller.rb
@@ -1,0 +1,6 @@
+class HotelsController < ApplicationController
+  def show
+    @hotel = Hotel.find(params[:id])
+    @guests = @hotel.guest_list
+  end
+end
--- a/app/models/guest.rb
+++ b/app/models/guest.rb
@@ -1,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+end
--- a/app/models/guest_room.rb
+++ b/app/models/guest_room.rb
@@ -1,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
--- a/app/models/hotel.rb
+++ b/app/models/hotel.rb
@@ -1,3 +1,8 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+  has_many :guests, through: :rooms
+
+  def guest_list
+    guests.distinct
+  end
 end
--- a/app/models/hotel.rb
+++ b/app/models/hotel.rb
@@ -3,6 +3,8 @@ class Hotel < ApplicationRecord
   has_many :guests, through: :rooms
 
   def guest_list
+    # I also googled this eleventeen thousand times to ensure it's active record but
+    # TBD...
     guests.distinct
   end
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -2,4 +2,10 @@ class Room < ApplicationRecord
   belongs_to :hotel
   has_many :guest_rooms
   has_many :guests, through: :guest_rooms
+
+  def guest_count
+    # Thorough googling tells me this is an active record method and 
+    # I hope to suite (pun intended) baby jiminy cricket they're correct
+    guests.count
+  end
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
   belongs_to :hotel
+  has_many :guest_rooms
+  has_many :guests, through: :guest_rooms
 end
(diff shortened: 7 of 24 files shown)


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestRoomsController < ApplicationController
+  def create
+    @guest_room = GuestRoom.create!(guest_id: params[:id], room_id: params[:room_id])
+    redirect_to "/guests/#{@guest_room.guest_id}"
+  end
```
megstang, line 5: This looks good. I like that you're directly creating the `GuestRoom` object rather than finding the room, finding the guest, and THEN making the association. I'm curious if you need line 3 stored in an instance variable since you don't have a view associated with this create action. Similarly, if you chose to get rid of that instance variable, you could just call `params[:id]` on line 4 when trying to interpolate the guest room.

megstang: One thing to be careful of with the `!` on line 3 is that your code will throw a gross user-side error (on the browser) if for whatever reason that GuestRoom couldn't be created (maybe they went in an invalid room id or something). Those bangs (`!`) can be super helpful in tests, OR they are helpful when you have exception handlers. If you're interested, take a look at exception handlers to see how you can utilize forced errors in your code.

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
+  end
```
megstang, line 5: We could remove line 4 and just pass `@guest` to the view since the guest object knows about the rooms associated with it. We want to send as few variables to the view as we can, so always consider "Is this information available with any other objects I'm sending off to the view?"

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,6 @@
+class HotelsController < ApplicationController
+  def show
+    @hotel = Hotel.find(params[:id])
+    @guests = @hotel.guest_list
+  end
+end
```
megstang, line 6: Similar to my above comment, you can call `guest_list` directly on the Hotel object in your view. So, no need to pass over a new variable.

app/models/guest.rb
```diff
@@ -0,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+end
```

app/models/guest_room.rb
```diff
@@ -0,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
```

app/models/hotel.rb
```diff
@@ -1,3 +1,10 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+  has_many :guests, through: :rooms
+
+  def guest_list
+    # I also googled this eleventeen thousand times to ensure it's active record but
+    # TBD...
+    guests.distinct
+  end
```
megstang, line 9: Hah yes... it is indeed an AR method ;) Nice work. And, good job pulling this into a model method in your Hotel class. Since it's data logic, this is exactly where it should go.

app/models/room.rb
```diff
@@ -1,3 +1,11 @@
...
   belongs_to :hotel
+  has_many :guest_rooms
+  has_many :guests, through: :guest_rooms
+
+  def guest_count
+    # Thorough googling tells me this is an active record method and 
+    # I hope to suite (pun intended) baby jiminy cricket they're correct
+    guests.count
+  end
 end
```
megstang, line 11: Great work :)

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,24 @@
...
+<% end %>
+</div>
+
+<div id="Add_room">
+<%= form_with url: "/guests/#{@guest.id}/rooms", method: :post do |form| %>
+    <p> <%= form.label :room_id, "Please enter a valid room ID:" %> </p>
+    <p> <%= form.text_area :room_id %> </p>
+    <p> <%= form.submit "Submit" %> </p>
+<% end %>
+</div>
```
megstang, line 24: This all looks good! One thing we could do to make your route on line 19 more RESTful is to use the route 'POST /guest_rooms'. However, that would require that you use a hidden field in your form to pass through the guest id within the body of the request (along with the room_id).

app/views/hotels/show.html.erb
```diff
@@ -0,0 +1,13 @@
...
+<h2>Location: <%= @hotel.location %></h2>
+</div>
+
+<div id="Guest_list">
+<h3> List of guests that have stayed at this hotel: </h3>
+<% @guests.each do |guest| %>
+  <p><%= guest.name %></p>
+  <p>------------------------------</p>
+<% end %>
+</div>
```
megstang, line 13: Nice! Following through with my comment from your HotelController, you could call `@hotel.guest_list` on line 9, and then iterate through that return value.

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,12 @@
...
+<% @rooms.each do |room| %>
+<div id="<%= room.suite %>">
+  <p>Suite: <%= room.suite %></p>
+  <p>Rate: <%= room.rate %></p>
+  <p>Hotel: <%= room.hotel.name %></p>
+  <p>Number of guests hosted: <%= room.guest_count %></p>
+  <p>------------------------------</p>
+</div>
+<% end %>
+</div>
```
megstang, line 12: 👏

config/routes.rb
```diff
@@ -3,4 +3,11 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/guests/:id", to: "guests#show"
+  post "/guests/:id/rooms", to: "guest_rooms#create"
```
megstang, line 8: following through with my comment on your form earlier, we could change this route to be: 

```ruby
post "/guest_rooms", to: "guest_rooms#create"
```

megstang (commented):
@ParkerBoeing Nice work! Let me know if you have any questions on my comments. I didn't leave any notes on testing because they were all looking really good -- nice job remembering to test Model methods, and great implementation of using classes/id's to target your testing with 'within'. I'll follow up in slack with more feedback by EOD today.

PR #2, opened 2023-07-24
Adrian Roussev

adrianRoussev:
completed user stories 1-3 and extension

### DIFF
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,0 +1,13 @@
+
+class GuestRoomsController < ApplicationController
+    def create
+        @guest = Guest.find(params[:id])
+        @room = Room.find(params[:guest_room][:room_id])
+        
+        # Create the association between the guest and the room
+        @guest.rooms << @room
+      
+      redirect_to "/guest/#{@guest.id}", notice: 'Room was successfully added to the guest.'
+    end
+  end
+  
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,0 +1,7 @@
+class GuestsController < ApplicationController
+    def show
+      @guest = Guest.find(params[:id])
+      @guest_rooms = @guest.rooms
+    end
+  end
+  
--- a/app/controllers/hotels_controller.rb
+++ b/app/controllers/hotels_controller.rb
@@ -1,0 +1,8 @@
+
+class HotelsController < ApplicationController
+    def show
+      @hotel = Hotel.find(params[:id])
+      @unique_guests = @hotel.guests.distinct
+    end
+  end
+  
--- a/app/controllers/rooms_controller.rb
+++ b/app/controllers/rooms_controller.rb
@@ -1,0 +1,7 @@
+
+class RoomsController < ApplicationController
+    def index
+      @rooms = Room.includes(:hotel, :guests).all
+    end
+  end
+  
--- a/app/models/guest.rb
+++ b/app/models/guest.rb
@@ -1,0 +1,8 @@
+class Guest < ApplicationRecord
+    validates :name, presence: true
+    validates :nights, presence: true, numericality: true
+    has_many :guest_rooms
+    has_many :rooms, through: :guest_rooms
+    has_many :hotels, through: :rooms
+  end
+  
--- a/app/models/guest_room.rb
+++ b/app/models/guest_room.rb
@@ -1,0 +1,5 @@
+class GuestRoom < ApplicationRecord
+    belongs_to :guest
+    belongs_to :room
+  end
+  
--- a/app/models/hotel.rb
+++ b/app/models/hotel.rb
@@ -1,3 +1,4 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+  has_many :guests, through: :rooms
 end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,20 @@
+
+<h1><%= @guest.name %></h1>
+
+<h2>Rooms they've stayed in:</h2>
+<ul>
+  <% @guest.rooms.each do |room| %>
+    <li>
+      Suite: <%= room.suite %><br>
+      Nightly rate: <%= room.rate %><br>
+      Hotel: <%= room.hotel.name %>
+    </li><br/>
+  <% end %>
+</ul>
+
+<h2>Add a room:</h2>
+<%= form_with model: [@guest, GuestRoom.new], url: "/guest/#{@guest.id}/add_room" do |f| %>
+  <%= f.label :room_id, 'Room ID:' %>
+  <%= f.number_field :room_id %>
+  <%= f.submit "Add Room" %>
+<% end %>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -14,7 +14,7 @@
 
 <h2>Add a room:</h2>
 <%= form_with model: [@guest, GuestRoom.new], url: "/guest/#{@guest.id}/add_room" do |f| %>
-  <%= f.label :room_id, 'Room ID:' %>
+  <%= f.label :room_id, 'Room:' %>
   <%= f.number_field :room_id %>
   <%= f.submit "Add Room" %>
 <% end %>
--- a/app/views/rooms/index.html.erb
+++ b/app/views/rooms/index.html.erb
@@ -1,0 +1,14 @@
+
+<h1>Rooms Index</h1>
+
+<ul>
+  <% @rooms.each do |room| %>
+    <li>
+      Suite: <%= room.suite %><br>
+      Nightly rate: <%= room.rate %><br>
+      Hotel: <%= room.hotel.name %><br>
+      Number of Guests: <%= room.guests.count %>
+    </li><br/>
+
+  <% end %>
+</ul>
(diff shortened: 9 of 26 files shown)


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,13 @@
+
+class GuestRoomsController < ApplicationController
+    def create
+        @guest = Guest.find(params[:id])
+        @room = Room.find(params[:guest_room][:room_id])
+        
+        # Create the association between the guest and the room
+        @guest.rooms << @room
+      
+      redirect_to "/guest/#{@guest.id}", notice: 'Room was successfully added to the guest.'
```
cjsim89, line 10: Careful of your routes - the convention is to keep routes plural, so: `"/guests/...`

cjsim89: This can be kinda misleading when working with a show page (we think of it as singular, but the route names should still be plural)

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,7 @@
+class GuestsController < ApplicationController
+    def show
+      @guest = Guest.find(params[:id])
+      @guest_rooms = @guest.rooms
```
cjsim89, line 4: This is fine to put in the controller, but we can still call `@guest.rooms` in the view.

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,8 @@
+
+class HotelsController < ApplicationController
+    def show
+      @hotel = Hotel.find(params[:id])
+      @unique_guests = @hotel.guests.distinct
```
cjsim89, line 5: Doing `.distinct` in the controller is an MVC infraction, since it changes the query from the DB. As such, it should belong in a model method and then tested.

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,7 @@
+
+class RoomsController < ApplicationController
+    def index
+      @rooms = Room.includes(:hotel, :guests).all
```
cjsim89, line 4: I think maybe doing `.includes` is fine, but we wouldn't need a `.all` afterward, since these two methods sorta contradict (or overlap) each other. I would double-check that you understand the difference here, and if it's unclear, just keep using `.all`. We don't require you to know `.includes` for at least mod 2.

app/models/guest.rb
```diff
@@ -0,0 +1,8 @@
+class Guest < ApplicationRecord
+    validates :name, presence: true
+    validates :nights, presence: true, numericality: true
```
cjsim89, line 3: Good validations! Just know for the future that we won't require validations on ICs other than if you want to personally practice them.

app/models/guest_room.rb
```diff
@@ -0,0 +1,5 @@
+class GuestRoom < ApplicationRecord
+    belongs_to :guest
+    belongs_to :room
```
cjsim89, line 3: Good job setting up your many-to-many relationship! The only thing missing is your model test for this model. Remember that each model and each line in that model must always be tested, even if it's just two `belongs_to` lines like this.

app/models/hotel.rb
```diff
@@ -1,3 +1,4 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+  has_many :guests, through: :rooms
```
cjsim89, line 3: Nice job with this association! 🙌

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    <li>
+      Suite: <%= room.suite %><br>
+      Nightly rate: <%= room.rate %><br>
+      Hotel: <%= room.hotel.name %>
+    </li><br/>
+  <% end %>
+</ul>
+
+<h2>Add a room:</h2>
+<%= form_with model: [@guest, GuestRoom.new], url: "/guest/#{@guest.id}/add_room" do |f| %>
```
cjsim89, line 16: I think it's cool you're going after the use of `form_with :model` syntax, but using it in this way makes it a bit more complicated than we need to. In class this week we'll learn how to use this without making an MVC infraction - the infraction I'm talking about is using `GuestRoom.new` in the view - this is something that the controller should do, and send this new instance down to the controller. 

In addition, the `add_room` part of the end of the URL is not considered RESTful since it's outside of the 7 main actions: `index, new, create, show, edit, update, delete`. However, you route it to a joined (`guest_rooms_controller`) which is the right approach.

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,14 @@
+
+<h1>Rooms Index</h1>
+
+<ul>
+  <% @rooms.each do |room| %>
+    <li>
+      Suite: <%= room.suite %><br>
+      Nightly rate: <%= room.rate %><br>
+      Hotel: <%= room.hotel.name %><br>
+      Number of Guests: <%= room.guests.count %>
```
cjsim89, line 10: Doing `.count` in this context is an ActiveRecord method, and so does not belong in the view, but rather in a model method and then tested. This is considered an MVC infraction.

config/routes.rb
```diff
@@ -1,6 +1,13 @@
...
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/guest/:id", to: "guests#show", as: :guest
+
+  post "/guest/:id/add_room", to: "guest_rooms#create", as: :add_guest_room
+
+  get "/rooms", to: "rooms#index", as: :rooms
+
+  get "/hotels/:id", to: "hotels#show", as: :hotel
```
cjsim89, line 11: Routes overall look good (except for the `post` which I talked about earlier). I like seeing you use these aliases; when you explore the Advanced Routing class you'll see how Rails does this for us!

db/seeds.rb
```diff
@@ -5,3 +5,32 @@
...
+
+guest1 = Guest.create(name: 'Charlize Theron', nights: 3)
+guest2 = Guest.create(name: 'Leonardo DiCaprio', nights: 5)
+guest3 = Guest.create(name: 'Natalie Portman', nights: 2)
+
+
+GuestRoom.create(guest: guest1, room: room1)
+GuestRoom.create(guest: guest1, room: room3)
+GuestRoom.create(guest: guest2, room: room2)
+GuestRoom.create(guest: guest3, room: room4)
```
cjsim89, line 34: Nice job setting up your seeds file! Remember to use the bang operator (`!`) when doing `.create!` so that Rails will give us more pertinent errors in the console if something were to go wrong.

spec/features/guest_rooms/add_guest_to_room_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  it 'adds a room to the guest' do
+    visit "/guest/#{guest.id}"
+    new_room = Room.create(rate: 200, suite: 'Deluxe', hotel: hotel)
+
+    fill_in 'guest_room_room_id', with: new_room.id
+    click_button 'Add Room'
+
+    expect(page).to have_content(new_room.suite)
+    expect(page).to have_content(new_room.rate)
+    expect(page).to have_content(hotel.name)
```
cjsim89, line 18: This test setup is fine and makes sense, but I think it can be stronger if we test that this page _does not_ have this content first, so that when we check that it _does_ have this new content, it means more.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  it 'allows adding a room to the guest' do
+    new_room = Room.create(rate: 200, suite: 'Deluxe', hotel: hotel)
+
+    visit "/guest/#{guest.id}"
+    fill_in 'guest_room_room_id', with: new_room.id
+    click_button 'Add Room'
+
+    expect(page).to have_content(new_room.suite)
+    expect(page).to have_content(new_room.rate)
+    expect(page).to have_content(hotel.name)
```
cjsim89, line 31: I think this is the same or similar test for the other user story in the other spec file... But, ok!

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  it 'displays a list of all rooms with their details and guest count' do
+    visit "/rooms"
+    expect(page).to have_content(room1.suite)
+    expect(page).to have_content(room1.rate)
+    expect(page).to have_content(hotel1.name)
+    expect(page).to have_content(room1.guests.count)
+
+    expect(page).to have_content(room2.suite)
```
cjsim89, line 24: The test setup is fine, but I find it more useful to test using strings exactly as it will appear on the page. Otherwise, capybara will look _**anywhere**_ in the page document for each thing separately, regardless of the order in which the tests appear. Example: 
```
expect(page).to have_content("Stayed at Alpine Resort's Deluxe suite for $200/night 2 times")
```
this is a more accurate test because we're expecting all the data being asked for is on the page in this specific context. The test you currently have is asking "Hey capybara, is the number `2` present anywhere on the page?", which will pass if there's a timestamp on the page and it's 12:00. So, accurate feature tests do matter in the long run.

spec/models/hotel_spec.rb
```diff
@@ -3,5 +3,6 @@
 RSpec.describe Hotel, type: :model do
   describe "relationships" do
     it { should have_many :rooms }
+    it { should have_many(:guests).through(:rooms) }
```
cjsim89, line 6: Good job adding this test! Don't forget to add a model test file for GuestRoom as well.

PR #4, opened 2023-07-24
Davis Weimer

Dav1s-Ops:
This was fun, I struggled a bit on creating/migrating the associations. Once those were created, however, it was very cool to be able to access all the instances through their dot notation (ex: @hyatt.rooms.first.guests).  Rails is very magical! Hopefully, I followed MVC convention well. I made it through all the User Stories including the extension.

### DIFF
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,0 +1,11 @@
+class GuestRoomsController < ApplicationController
+  def create
+    guest = Guest.find(params[:id])
+    room = Room.find(params[:add_room_id])
+
+    guest.rooms.push(room)
+    guest.save
+
+    redirect_to "/guests/#{guest.id}"
+  end
+end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -6,4 +6,8 @@ class Room < ApplicationRecord
   def guest_count
     guests.count
   end
+
+  def guests_distinct
+    guests.distinct
+  end
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
   belongs_to :hotel
+  has_many :guest_rooms
+  has_many :guests, through: :guest_rooms
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -2,4 +2,8 @@ class Room < ApplicationRecord
   belongs_to :hotel
   has_many :guest_rooms
   has_many :guests, through: :guest_rooms
+
+  def guest_count
+    guests.count
+  end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,4 +3,6 @@ Rails.application.routes.draw do
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/guests/:id", to: "guests#show"
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,4 +6,6 @@ Rails.application.routes.draw do
 
   get "/guests/:id", to: "guests#show"
   post "/guests/:id", to: "guest_rooms#create"
+
+  get "/rooms", to: "rooms#index"
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,4 +5,5 @@ Rails.application.routes.draw do
   # root "articles#index"
 
   get "/guests/:id", to: "guests#show"
+  post "/guests/:id", to: "guest_rooms#create"
 end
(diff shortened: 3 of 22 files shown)


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,11 @@
+class GuestRoomsController < ApplicationController
+  def create
+    guest = Guest.find(params[:id])
+    room = Room.find(params[:add_room_id])
+
+    guest.rooms.push(room)
+    guest.save
```
juliet-e, line 7: I believe (but it might be worth checking) that if you add a room to the guest's room collection, it should save automatically. But I could be wrong! Under the hood, when you add to the rooms collection, Rails is creating a joins record (GuestRoom.create(room_id: room_id, guest_id: guest_id)) rather than just instantiating a joins record. Will double check this

app/models/room.rb
```diff
@@ -1,3 +1,13 @@
 class Room < ApplicationRecord
   belongs_to :hotel
+  has_many :guest_rooms
+  has_many :guests, through: :guest_rooms
+
+  def guest_count
+    guests.count
+  end
+
+  def guests_distinct
```
juliet-e, line 10: This works just fine, but another option that is also VERY magical: putting this method in the Hotel model
You might be thinking.. how would the hotel model have an attribute called `guests`? And it wouldn't but you could add an additional association to hotel to allow it mooch off the associations set up for rooms. Rooms belong to hotels, and rooms has the association to guests, so if you add `has_many :guests, through: :rooms` to the hotel model, you can actually do `hotel.guests`! It's magic! Try it out

config/routes.rb
```diff
@@ -3,4 +3,11 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/guests/:id", to: "guests#show"
+  post "/guests/:id", to: "guest_rooms#create"
```
juliet-e, line 8: A more RESTful option here would be "/guests/:id/rooms" to indicate that you're posting to a path that includes both guests and rooms (it's a joins record, not just a guest record). Otherwise, this path makes it seem like it's just creating another guest. I like that you created a joins controller though 👍

spec/spec_helper.rb
```diff
@@ -11,7 +11,25 @@
 # a separate helper file that requires the additional dependencies and performs
 # the additional setup, and require it from the spec files that actually need
 # it.
-#
+def test_data
```
juliet-e, line 14: Nice use of spec_helper to centralize test data!

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  describe "#show page" do
+    it "displays the guests name and all info about the rooms stayed in" do
+      @billy.rooms.push(@motel_8_room_1)
+      @billy.rooms.push(@daysinn_room_5)
+
+      visit "/guests/#{@billy.id}"
+
+      expect(page).to have_content(@billy.name)
+
+      expect(page).to have_content(@billy.rooms.first.hotel.name)
```
juliet-e, line 16: `within` blocks would help make these assertions more robust, but at the very least, it's good to include the fixed text around the dynamic values to make sure the value is in the right place, i.e. `have_content("Room Rate: #{room.rate}")`

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      visit "/rooms"
+
+      expect(page).to have_content(@motel_8_room_1.hotel.name)
+      expect(page).to have_content(@motel_8_room_1.rate)
+      expect(page).to have_content(@motel_8_room_1.suite)
+
+      expect(page).to have_content(@hyatt_room_3.hotel.name)
+      expect(page).to have_content(@hyatt_room_3.rate)
+      expect(page).to have_content(@hyatt_room_3.suite)
+      expect(page).to have_content(@hyatt_room_3.guest_count)
```
juliet-e, line 27: Same note here. Once this variable resolves, you're just asserting that a small integer appears on the page. Making these assertions more explicit with fixed text and within blocks would make the test more meaningful

spec/models/room_spec.rb
```diff
@@ -1,7 +1,36 @@
...
+    test_data
+  end
   describe "relationships" do
     it { should belong_to :hotel }
+    it { should have_many(:guest_rooms) }
+    it { should have_many(:guests).through(:guest_rooms) }
+  end
+
+  describe "instance methods" do
+    it "#guest_count" do
```
juliet-e, line 14: Nice work testing the model methods!

PR #5, opened 2023-07-24
Ian Lyell PR

ILyell:


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestRoomsController < ApplicationController
+    def create
+        @guest_room = GuestRoom.create(room_id: params[:room_id], guest_id: params[:id])
+        redirect_to "/guest/#{params[:id]}"
+    end
+end
```
megstang, line 6: This looks great! I like that you're creating your GuestRoom right away with the info you pass in.

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestsController < ApplicationController
+    
+    def show
+        @guest = Guest.find(params[:id])
+    end
+end
```
megstang, line 6: 👍

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RoomsController < ApplicationController
+
+    def index
+        @rooms = Room.all
+    end
+end
```
megstang, line 6: 👍

app/models/room.rb
```diff
@@ -1,3 +1,9 @@
 class Room < ApplicationRecord
   belongs_to :hotel
+  has_many :guest_rooms
+  has_many :guest, through: :guest_rooms
+
+  def guest_count
+    guest_rooms.count
+  end
```
megstang, line 8: Nice use of a class method here

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    <br>
+<% end %>
+</section>
+<section id="room_form">
+<%= form_with url: "/guest/#{@guest.id}", method: :post, local: true do |form| %>
+    <br>
+    <%= form.label :room_id %>
+    <%= form.text_field :room_id %>
+    <%= form.submit "Add Room" %>
+<% end %>
```
megstang, line 20: Great work! Your route on line 15 could be a bit more restful since you're actually creating a GuestRoom. It could go to something like '/guest_rooms' and pass the guest id via a hidden field in the form. Or, it might even make more sense to have something like "/guests/:id/rooms"

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+                    visit "guest/#{@guest_1.id}"
+
+                    within "#guest_info" do
+                        expect(page).to have_content(@guest_1.name)
+                        expect(page).to have_content(@room_1.suite)
+                        expect(page).to have_content(@room_1.rate)
+                        expect(page).to have_content(@hotel_1.name)
+                    end
+
+                end
```
megstang, line 41: Since you're testing that all rooms will appear on this page, it might be nice to have more than one room for guest_1 to make sure that it can iterate through a collection an display ALL.

config/routes.rb
```diff
@@ -3,4 +3,12 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  # Guest
+  get "guest/:id", to: "guests#show"
+  post "guest/:id", to: "guest_rooms#create"
```
megstang, line 9: We want our routs to be named similarly to their resources (plural). So, something like '/guests/:id'

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+                within "#room_#{@room_3.id}" do
+                    expect(page).to have_content(@room_3.suite)
+                    expect(page).to have_content(@room_3.rate)
+                    expect(page).to have_content(@room_3.hotel.name)
+                    expect(page).to have_content("Guest: 1")
+                end
+            end
+        end
+    end
+end
```
megstang, line 56: great use of `within`s!

megstang (commented):
@ILyell Nice work! I'll follow up in slack by EOD with more feedback.

PR #6, opened 2023-07-24
Chris C

topher-nullset:
this was fun, i was expecting more model methods.


app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
```
juliet-e, line 4: Keep in mind that you don't necessarily need to create the @rooms instance var as you can reference it in the view as @guest.rooms. There are Rails-ists who argue that minimizing instance variables is best practice to make sure that your controllers are adhering to SRP principles (Sandi Metz is one of these Rails experts)

This is a for-fun comment, not affecting your grade

app/controllers/room_guests_controller.rb
```diff
@@ -0,0 +1,9 @@
+class RoomGuestsController < ApplicationController
+  def create
+    guest = Guest.find(params[:id])
+    room = Room.find(params[:room_id])
+    RoomGuest.create!(guest_id: guest.id, room_id: room.id)
```
juliet-e, line 5: We are not expecting much error handling in mid mods, but keep in mind that in general, it's best to use `create!` only in tests or seeds, and in controllers just use `create` so that users are not exposed to unexpected exceptions being thrown.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  it "has a form to add a new room for this guest" do
+    visit "/guests/#{@guest1.id}"
+    # save_and_open_page
+    expect(page).to have_content("Add a Room")
+    fill_in "room_id", with: @room2.id
+    click_button "Submit"
+    # save_and_open_page
+    expect(current_path).to eq("/guests/#{@guest1.id}")
+    expect(page).to have_content("Charlize Theron")
+  end
```
juliet-e, line 34: These tests don't cover the functionality of listing room history - make sure you add some assertions that verify the presence of room information. You've got your view set up so that you can use within blocks here to ensure the placement of the room information is correct. 

In the form test case, I'd also expect to see an assertion around the presence of the newly added room information

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    RoomGuest.create!(guest_id: @guest1.id, room_id: @room2.id)
+    RoomGuest.create!(guest_id: @guest1.id, room_id: @room4.id)
+    RoomGuest.create!(guest_id: @guest2.id, room_id: @room1.id)
+    RoomGuest.create!(guest_id: @guest2.id, room_id: @room4.id)
+  end
+
+  it "shows list of all rooms and information about each room" do
+    visit "/rooms"
+    # save_and_open_page
+    within("#room-#{@room1.id}") do
```
juliet-e, line 22: This is a great, robust test!

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1>Room Index</h1>
+<% @rooms.each do |room|%>
+  <div id="room-<%= room.id %>">
+    Room#: <%= room.id %> --- 
+    Suite: <%= room.suite%> ---
+    Hotel: <%= room.hotel.name %> ---
+    Room Guests: <%= room.guests.count %>
```
juliet-e, line 7: We'd rather you encapsulate the guest counting to a model method in the Room model for a couple reasons
1) As a general rule, it's a calculation (although minor) so it should remain close to the data rather than living in presentation logic
2) The counting could be more complex. For example, you could include `distinct` or other conditions. Encapsulating it in the model allows you to test it easily in a more isolated environment. 

The rubric allows for one MVC infraction so this doesn't affect your score in that category

app/controllers/guests_controller.rb
topher-nullset: In that case would I need to join the tables in the model instead of using .find in the controller?

juliet-e: Nope, in your view whenever you need to access rooms, you would just access them using the `@guest` instance variable
i.e. <% @guest.rooms.each %>

Line 3 here should remain the same - you retrieve the guest by the ID

PR #7, opened 2023-07-24
Jorja

JorjaF:
* A reflection on how you felt you did with this challenge
I guess i did okay on this, i hate having to start with a join table

* What story you got through
I got thru story 3. i was nervous to try the extensions because that is why I failed last time and i didn't want that to happen again

### DIFF
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -2,5 +2,11 @@ class GuestsController < ApplicationController
   def show
     @guest = Guest.find(params[:id])
   end
+
+  def add_room
+    @guest = Guest.find(params[:id])
+    @guest.rooms << Room.find(params[:room_id])
+    redirect_to "/guest/#{@guest.id}"
+  end
   
 end
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,0 +1,6 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+  end
+  
+end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,7 @@
+name: <%= @guest.name %>
+hotels: <% @guest.rooms.each do |room|%>
+  suite: <%= room.suite %>
+  rate: <%= room.rate %>
+  hotel: <%= room.hotel.name %>
+
+<%end%>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -5,3 +5,9 @@ hotels: <% @guest.rooms.each do |room|%>
   hotel: <%= room.hotel.name %>
 
 <%end%>
+
+<%= form_with url: "/guest/#{@guest.id}/add_room" do |f| %>
+  <%= f.label :room_id %>
+  <%= f.text_field :room_id %>
+  <%= f.submit "Add Room" %>
+<% end %>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,13 +1,15 @@
-name: <%= @guest.name %>
-hotels: <% @guest.rooms.each do |room|%>
-  suite: <%= room.suite %>
-  rate: <%= room.rate %>
-  hotel: <%= room.hotel.name %>
+Name: <%= @guest.name %> <br>
+Hotels: <% @guest.rooms.each do |room|%> 
+  <br>
+  Hotel: <%= room.hotel.name %>
+  Suite: <%= room.suite %> 
+  Rate: <%= room.rate %>
+  
 
 <%end%>
-
+<br>
+Add a room to this guest:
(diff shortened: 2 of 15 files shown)


app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,12 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+  end
+
+  def add_room
```
cjsim89, line 6: The content of this method is correct, but I wouldn't use an action called `add_room` since it's not part of the 7 RESTful routes (`index, show, new, create, edit, update, delete`); rather what we're doing in this method is _creating_ an instance of a `Booking` model that you made. So, a better name for this action would be `create`. 

Also, since we're creating a Booking and not a Guest, it really belongs in a `BookingController` file. But once that's created, we could just lift the code in this method right over there, and it would work great there too.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <br>
+  Hotel: <%= room.hotel.name %>
+  Suite: <%= room.suite %> 
+  Rate: <%= room.rate %>
+  
+
+<%end%>
+<br>
+Add a room to this guest:
+<%= form_with url: "/guest/#{@guest.id}/add_room" do |f| %>
```
cjsim89, line 12: Good job with the form setup overall; don't forget your `method: :post` and `local: true` as well.

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<% @rooms.each do |room| %>
+  <p>Suite: <%= room.suite %></p>
+  <p>Price: <%= room.rate %></p>
+  <p>Hotel: <%= room.hotel.name %></p>
+  <p>Number of Guests: <%= room.guests.count %></p>
```
cjsim89, line 5: Doing `.count` here is an MVC infraction, since `.count` changes the SQL that the database uses to find the number of guests for this room. As such, it belongs in the model as a new method, and then we should test that method in a model test. 

If you want to try it yourself, put a `binding.pry` here and type `room.guests.count.to_sql` - see what the SQL is returned as!

config/routes.rb
```diff
@@ -3,4 +3,11 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "guest/:id", to: "guests#show"
+  post "guest/:id/add_room", to: "guests#add_room"
```
cjsim89, line 8: These routes are RESTful overall, good job - I would change this route to reflect the comment I made above about where we should be creating a new instance of Booking. I would do something like: 

```
post "/guests/:id/rooms", to: "bookings#create"
```

Also, remember that the resources in our routes should always be plural; so `guests` just like you have `rooms` on line 10.

spec/features/guest/guest_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  # When I visit a guest's show page
+  # I see the guest's name
+  # And I see a list of all the rooms they've stayed in
+  # including the room's suite, nightly rate, and the name of the hotel that it belongs to.
+  
+  it "can see guest's name and a list of rooms they have stayed in" do
+    room_2 = Room.create!(rate: 200, suite: "The Red Room", hotel_id: hotel_1.id)
+    guest_1.rooms << room_1
+    guest_1.rooms << room_2
+    guest_1.save
```
cjsim89, line 18: This `.save` is not required in a feature test; you can put it in a Controller though, if you're updating or creating a new instance of something.

spec/features/room/room_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    visit "/rooms"
+
+    expect(page).to have_content(room_1.suite)
+    expect(page).to have_content(room_1.rate)
+    expect(page).to have_content(hotel_1.name)
+    expect(page).to have_content(room_2.suite)
+    expect(page).to have_content(room_2.rate)
+    expect(page).to have_content(room_1.guests.count)
+    expect(page).to have_content(room_2.guests.count)
```
cjsim89, line 27: Your feature tests look good overall! 

The only thing missing from your spec files are your additional **model tests**. If we follow TDD, anytime we make a new model, or add a new relationship to a model, or create a new method in any model, we need to be sure we test it _first_. Remember to always have your model tests fully covered at any point through building your applications.

spec/features/guest/guest_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    visit "/guest/#{guest_1.id}"
+
+    expect(page).to have_content(guest_1.name)
+        
+    fill_in :room_id, with: room_1.id
+    click_button "Add Room"
+
+    expect(current_path).to eq("/guest/#{guest_1.id}")
+    expect(page).to have_content(guest_1.name)
+    expect(page).to have_content(room_1.suite)
```
cjsim89, line 52: This test is good; one way to improve it would be to test that `room_1.suite` does _not_ appear on the page before clicking the "Add Room" button. So, above that line I would put 

```
expect(page).to_not have_content(room_1.suite)
```

PR #8, opened 2023-07-24
Chris Asercion Mid Mod PR

ChrisAsercion:
I think I did pretty well on this one. I got stuck around User Story 2 but used resources to get through it. I was fairly surprised how far I got. I managed to catch myself from using `.count` in the view.

### DIFF
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,0 +1,11 @@
+class GuestRoomsController < ApplicationController
+  def create
+    guest = Guest.find(params[:id])
+    room = Room.find(params[:room_id])
+
+    guest.rooms << room
+    guest.save
+
+    redirect_to "/guests/#{guest.id}" 
+  end
+end
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,0 +1,17 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+  end
+
+  # def update
+  #   guest = Guest.find(params[:id])
+  #   #require 'pry'; binding.pry
+  #   guest.add_room
+  #   guest.update
+  #   redirect_to "guest/#{guest.id}"
+  # end
+
+  # def new_room_params
+  #   params.permit(:room_id)
+  # end
+end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -2,4 +2,9 @@ class Room < ApplicationRecord
   belongs_to :hotel
   has_many :guest_rooms
   has_many :guests, through: :guest_rooms
+  
+  def number_of_guests
+    self.guests.count
+  end
+  
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
   belongs_to :hotel
+  has_many :guest_rooms
+  has_many :guests, through: :guest_rooms
 end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -2,8 +2,9 @@
 <p>rooms:</p>
 <ul>
   <% @guest.rooms.each do |room| %>
-  <li><%= room.rate %>$ per night</li>
+  <p></p>
   <li><%= room.suite %></li>
+  <li><%= room.rate %>$ per night</li>
   <li><%= room.hotel.name %> hotel</li>
   <% end %>
 </ul>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,17 @@
+<h3><%= @guest.name %></h3>
+<p>rooms:</p>
+<ul>
+  <% @guest.rooms.each do |room| %>
+  <li><%= room.rate %>$ per night</li>
+  <li><%= room.suite %></li>
+  <li><%= room.hotel.name %> hotel</li>
+  <% end %>
+</ul>
+
+<%= form_with url: "/guests/#{@guest.id}", method: :patch, id: 'Add Room' do |f| %>
+  <h3>Select room:</h3>
+  <%= f.label :room_id %>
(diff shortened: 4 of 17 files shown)


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+  def create
+    guest = Guest.find(params[:id])
+    room = Room.find(params[:room_id])
+
+    guest.rooms << room
+    guest.save
+
+    redirect_to "/guests/#{guest.id}" 
+  end
+end
```
megstang, line 11: This looks good. One thing I really like about this class is how you've created your own controller for creation of a GuestRoom. 

Although your code here is clean and will work, I think there might be a more optimal way to do this. Currently, you are finding the Guest object AND the Room object, and then creating that connection between the guest and room. Instead, you have EVERYTHING you need right away to create that connection. You could have something like this to make your code more optimal:

```ruby
def create 
  GuestRoom.create(room_id: params[:room_id], guest_id: params[:id])
end 
```

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  #   #require 'pry'; binding.pry
+  #   guest.add_room
+  #   guest.update
+  #   redirect_to "guest/#{guest.id}"
+  # end
+
+  # def new_room_params
+  #   params.permit(:room_id)
+  # end
+end
```
megstang, line 17: 👍

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,18 @@
...
+<p>rooms:</p>
+<ul>
+  <% @guest.rooms.each do |room| %>
+  <p></p>
+  <li><%= room.suite %></li>
+  <li><%= room.rate %>$ per night</li>
+  <li><%= room.hotel.name %> hotel</li>
+  <% end %>
+</ul>
+
```
megstang, line 11: Looks great for US 1!

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,18 @@
...
+<%= form_with url: "/guests/#{@guest.id}", method: :patch, id: 'Add Room' do |f| %>
+  <h3>Select room:</h3>
+  <%= f.label :room_id %>
+  <%= f.text_area :room_id%>
+  <p>Submit this room: <%= f.submit "Submit"%></p>
+  </form>
+<% end %>
```
megstang, line 18: This form looks good, however, since in your routes, you have `PATCH /guests/:id` going to the `guest_rooms#create` action, this might make more sense to have this go to a `POST` since you're creating a new `GuestRoom` in the database. It could look something like `POST /guest_rooms, to: 'guest_rooms#create'`. And then, you'll need to pass in your `guest_id` and `room_id` via your form (hidden fields might help you pass your guest id through).

app/models/room.rb
```diff
@@ -1,3 +1,10 @@
 class Room < ApplicationRecord
   belongs_to :hotel
+  has_many :guest_rooms
+  has_many :guests, through: :guest_rooms
+  
+  def number_of_guests
+    self.guests.count
+  end
+  
 end
```
megstang, line 10: Nice job with this model method!

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,11 @@
...
+<p>rooms:</p>
+<ul>
+  <% @rooms.each do |room| %>
+  <li><%= room.rate %>$ per night</li>
+  <li><%= room.suite %></li>
+  <li><%= room.hotel.name %> hotel</li>
+  <li><%= room.number_of_guests %> guests have stayed in this room</li>
+  <p></p>
+  <% end %>
+</ul>
```
megstang, line 11: 👍

config/routes.rb
```diff
@@ -3,4 +3,8 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/guests/:id", to: "guests#show"
+  patch "/guests/:id", to: "guest_rooms#create"
```
megstang, line 8: reference my comment on your guest show page to see how I'd update this route ;)

spec/features/guest/show_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    it "displays a form to add a room to this guest" do
+      visit "/guests/#{@guest_1.id}"
+      #save_and_open_page
+      fill_in('room_id', with: @room_3.id)
+      
+      click_on 'Submit'
+      expect(page).to have_content("1 bedroom")
+      expect(page).to have_content(25)
+      #save_and_open_page
+    end
```
megstang, line 49: let's make lines 46 a bit more dynamic in case any of your test setup changes. Instead of hardcoding in the values you're looking for, you could do something like: 

```ruby
expect(page).to have_content(@room_3.suite)
expect(page).to have_content(@room_3.rate)
```

spec/models/guest_rooms.rb
```diff
@@ -0,0 +1,9 @@
+require "rails_helper"
+
+RSpec.describe GuestRoom, type: :model do
+  it '#create' do
+    @hotel_1 = Hotel.create!(name: "Sheraton", location: "Denver")
+    @room_1 = @hotel_1.rooms.create!(rate: 50, suite: "2 bedroom")
+    @guest_1 = @room_1.guests.create!(name: "Steve", nights: 3)
+
+end
```
megstang, line 9: You don't need to test your `create` method. This is a method that comes from ActiveRecord. Testing that your feature that allows user to create a GuestRoom is the main thing you need to test for this feature.

megstang: We will want to have relationship tests in here with shoulda_matchers, though... and throughout the other models that you've added relationships to. We should test that `GuestRoom` belongs to guest and belongs to room, that `Room` has many guests through guest_room, and that `Guest` has many rooms through guest_room.

megstang (commented):
@ChrisAsercion Nice work! I left some comments for you to look over. I'll follow up in slack later today.

PR #10, opened 2023-07-24
Matt Lim

MatthewTLim:
Went very well was able to complete all stories plus extension!

### DIFF
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,0 +1,9 @@
+class GuestRoomsController < ApplicationController
+  def create
+    @guest = Guest.find(params[:id])
+    @room = Room.find(params[:room_id])
+    @guest.rooms << @room
+    @guest.save
+    redirect_to "/guests/#{@guest.id}"
+  end
+end
--- a/app/controllers/hotels_controller.rb
+++ b/app/controllers/hotels_controller.rb
@@ -1,0 +1,6 @@
+class HotelsController < ApplicationController
+  def show
+    @hotel = Hotel.find(params[:id])
+    @unique_guests = @hotel.list_unique_guests
+  end
+end
--- a/app/models/hotel.rb
+++ b/app/models/hotel.rb
@@ -1,3 +1,7 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+
+  def list_unique_guests
+    Guest.distinct.joins(:guest_rooms => :room).where("rooms.hotel_id = ?", self.id)
+  end
 end
--- a/spec/features/guest_spec.rb
+++ b/spec/features/guest_spec.rb
@@ -1,0 +1,12 @@
+require 'rails_helper'
+
+Rspec.describe Guest do
+  #Story 1
+  #Guest Show
+
+  #As a visitor
+  #When I visit a guest's show page
+  #I see the guest's name
+  #And I see a list of all the rooms they've stayed in
+  #including the room's suite, nightly rate, and the name of the hotel that it belongs to.
+end
--- a/spec/features/guest_spec.rb
+++ b/spec/features/guest_spec.rb
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-Rspec.describe Guest do
+RSpec.describe Guest do
   #Story 1
   #Guest Show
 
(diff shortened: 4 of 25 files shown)


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,9 @@
+class GuestRoomsController < ApplicationController
+  def create
+    @guest = Guest.find(params[:id])
+    @room = Room.find(params[:room_id])
+    @guest.rooms << @room
+    @guest.save
```
cjsim89, line 6: Nice job writing this method in a 3rd controller - it's exactly the approach I like to take as well. 👍

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,6 @@
+class HotelsController < ApplicationController
+  def show
+    @hotel = Hotel.find(params[:id])
+    @unique_guests = @hotel.list_unique_guests
```
cjsim89, line 4: This is fine but not necessary, since we can access the `.list_unique_guests` method from the view as well.

app/models/hotel.rb
```diff
@@ -1,3 +1,7 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+
+  def list_unique_guests
+    Guest.distinct.joins(:guest_rooms => :room).where("rooms.hotel_id = ?", self.id)
```
cjsim89, line 5: Good job with this query, it works great! 

There's a way to make it a bit easier to think through, though, in my opinion, and it's by using additional associations. 
Considering that a Hotel has many Rooms, and a Room has many Guests, then the way we draw a direct line from Hotel to Guests is _through Rooms_. Therefore, it logically follows that we can make this association: 
```
```

Then, our query turns into something much simpler. Try refactoring this query with the addition of this association, and let me know what you find!

spec/features/guest_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    @room_2 = Room.create!(rate: 200, suite: "Presedential", hotel: @hotel_1)
+    @room_3 = Room.create!(rate: 300, suite: "Regular", hotel: @hotel_1)
+    @guest_1 = Guest.create!(name: "Jack Torrance", nights: 2)
+    @guest_1.rooms << @room_1
+    @guest_1.rooms << @room_2
+
+
+    visit "/guests/#{@guest_1.id}"
+
+    expect(page).to_not have_content(@room_3.suite)
```
cjsim89, line 51: Really nice test setup through all of your tests - both objects and the things you're expecting, including this line for a "negative" case before the real test on line 57. Keep this up! 🙌

spec/features/hotel_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    @guest_1.rooms << @room_2
+    @guest_2.rooms << @room_2
+
+    @guest_1.rooms << @room_3
+    @guest_2.rooms << @room_3
+    @guest_3.rooms << @room_3
+
+    visit "/hotels/#{@hotel_1.id}"
+
+    expect(page).to have_content(@guest_1.name)
```
cjsim89, line 28: I think the only thing that I'd like to see more of in your feature tests is the use of `within` blocks, to get you used to using HTML identifiers in the views. Otherwise, good job! 

As for this particular test, the user story asks for `unique list of all guests`. So, to make sure your feature test is also looking for unique values, you can do this in a few different ways.. Here's the ways I like to do it: 

```
# or
expect(page).to have_content(@guest_1.name).once
```

This way, if the content appears either more or less than the times you're expecting, the tests will fail.

spec/models/hotel_spec.rb
```diff
@@ -4,4 +4,27 @@
...
+      @guest_1.rooms << @room_1
+
+      @guest_1.rooms << @room_2
+      @guest_2.rooms << @room_2
+
+      @guest_1.rooms << @room_3
+      @guest_2.rooms << @room_3
+      @guest_3.rooms << @room_3
+
+      expect(@hotel_1.list_unique_guests).to eq([@guest_1, @guest_2, @guest_3])
```
cjsim89, line 27: This is fine, but if you're testing an instance method I would create at least 2 instances of the thing you're testing, that way you can prove it works in at least 2 different ways. So for example, making a 2nd Hotel instance with _no_ rooms would mean it would return an empty array for this method.

PR #11, opened 2023-07-24
Ethan Black

ethanrossblack:
I thought this IC went very well for me! It was very similar to the practice IC, so I felt like comfortable and familiar with the user stories.

I was able to complete all the way through the extension.

### DIFF
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,5 +1,5 @@
 class GuestsController < ApplicationController
   def show
-    @guest = Guest.find(params[:id])
+    @guest = Guest.find(params[:id]) # thing
   end
 end
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,0 +1,5 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+  end
+end
--- a/app/controllers/room_guests_controller.rb
+++ b/app/controllers/room_guests_controller.rb
@@ -1,0 +1,12 @@
+class RoomGuestsController < ApplicationController
+  def create
+    room_guest = RoomGuest.create(room_guest_params)
+
+    redirect_to "/guests/#{room_guest_params[:guest_id]}"
+  end
+
+  private
+  def room_guest_params
+    params.permit(:room_id, :guest_id)
+  end
+end
--- a/app/models/hotel.rb
+++ b/app/models/hotel.rb
@@ -3,5 +3,7 @@ class Hotel < ApplicationRecord
   has_many :room_guests, through: :rooms
   has_many :guests, through: :room_guests
 
-  
+  def unique_guests
+    guests.distinct
+  end
 end
--- a/app/models/hotel.rb
+++ b/app/models/hotel.rb
@@ -1,3 +1,7 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+  has_many :room_guests, through: :rooms
+  has_many :guests, through: :room_guests
+
+  
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -2,4 +2,8 @@ class Room < ApplicationRecord
   belongs_to :hotel
   has_many :room_guests
   has_many :guests, through: :room_guests
+
+  def hotel_name
+    hotel.name
+  end
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -6,4 +6,8 @@ class Room < ApplicationRecord
   def hotel_name
     hotel.name
   end
+
+  def total_guests
+    guests.count
+  end
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
   belongs_to :hotel
+  has_many :room_guests
+  has_many :guests, through: :room_guests
 end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -19,6 +19,16 @@
   <% end %>
 </div>
 
+<div id="add-room">
+  <h3>Add Room</h3>
+  <p>Enter the ID of an eisting room.</p>
+  <%= form_with url: "/guests/#{@guest.id}/rooms", method: :post, local: true do |f| %>
+    <%= f.label :room_id, "Room ID" %>
+    <%= f.text_field :room_id %>
+    <%= f.submit "Submit" %>
+  <% end %>
+</div>
+
 
 <!-- Idea if I have time: Maybe put in a partial for displaying a hotel? -->
 
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,6 +1,24 @@
 <h1>Guest Show Page</h1>
 
-<div id="guest-info">
+<div id="guest_info">
   <h2>Guest Info</h2>
   <p><b>NAME: </b><%= @guest.name %></p>
 </div>
+
+<div id="guest_room_history">
+  <h2><%= @guest.name %>'s Room History</h2>
+  
+  <% @guest.rooms.each do |room| %>
+    <div id="room-<%= room.id %>">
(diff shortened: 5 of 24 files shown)


app/controllers/room_guests_controller.rb
```diff
@@ -0,0 +1,12 @@
+class RoomGuestsController < ApplicationController
+  def create
+    room_guest = RoomGuest.create(room_guest_params)
+
+    redirect_to "/guests/#{room_guest_params[:guest_id]}"
```
cjsim89, line 5: Nice job creating this 3rd controller - it's exactly the approach I like to take too.

app/models/hotel.rb
```diff
@@ -1,3 +1,9 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+  has_many :room_guests, through: :rooms
+  has_many :guests, through: :room_guests
```
cjsim89, line 4: This association works, but it's actually one more line than we really need... Consider that a Hotel has many Rooms, and a Room has many Guests. Therefore, we could write an association that follows that logic: 

```
```

app/models/hotel.rb
```diff
@@ -1,3 +1,9 @@
...
+
+  def unique_guests
+    guests.distinct
```
cjsim89, line 7: But either way you write your associations, this is a really good use of AR through associations!

app/models/room.rb
```diff
@@ -1,3 +1,13 @@
 class Room < ApplicationRecord
   belongs_to :hotel
+  has_many :room_guests
+  has_many :guests, through: :room_guests
+
+  def hotel_name
+    hotel.name
```
cjsim89, line 7: Nice job with both of these instance methods!

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,34 @@
...
+        <p><b>HOTEL: </b>$<%= room.hotel_name %></p>
+      </ul>
+    <div>
+  <% end %>
+</div>
+
+<div id="add-room">
+  <h3>Add Room</h3>
+  <p>Enter the ID of an eisting room.</p>
+  <%= form_with url: "/guests/#{@guest.id}/rooms", method: :post, local: true do |f| %>
```
cjsim89, line 25: Great job with this form!

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,34 @@
...
+    <%= f.label :room_id, "Room ID" %>
+    <%= f.text_field :room_id %>
+    <%= f.submit "Submit" %>
+  <% end %>
+</div>
+
+
+<!-- Idea if I have time: Maybe put in a partial for displaying a hotel? -->
```
cjsim89, line 33: This might make sense if it's repeated code; but I'm not sure where a hotel is repeated code here...? Either way it might be good if you want to practice it!

config/routes.rb
```diff
@@ -3,4 +3,11 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/guests/:id", to: "guests#show"
+  post "guests/:guest_id/rooms", to: "room_guests#create"
```
cjsim89, line 8: Good RESTful routes!

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+    visit "/guests/#{@guest1.id}"
+    
+    within("div#guest_info") do
+      expect(page).to have_content(@guest1.name)
+      expect(page).to_not have_content(@guest2.name)
+    end
+    
+    visit "/guests/#{@guest2.id}"
+    
+    within("div#guest_info") do
```
cjsim89, line 38: Good use of `within` blocks as well

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+
+  it "I see a guest's name" do
+    visit "/guests/#{@guest1.id}"
+    
+    within("div#guest_info") do
+      expect(page).to have_content(@guest1.name)
+      expect(page).to_not have_content(@guest2.name)
+    end
+    
+    visit "/guests/#{@guest2.id}"
```
cjsim89, line 36: This is a very thorough feature test- keep this up!

cjsim89: The things I see that make it thorough: 
* Using within blocks
* Using `.to` and `.to_not` lines inside within blocks
* Doing all of that on multiple pages

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+        expect(page).to have_content(@room2.suite)
+        expect(page).to have_content(@room2.rate)
+        expect(page).to have_content(@hotel.name)
+      end
+    
+      expect(page).to_not have_css("div#room-#{@room3.id}")
+    end
+  end
+  # =================
+  # END STORY 1 TESTS
```
cjsim89, line 101: So, as thorough as these tests are, there's _a lot_ of them 😅 

I'm a fan of thoroughness, but not at the sake of time or overextending. I would recommend going through these tests again and re-assessing what, at a base level, is necessary or required for _proving_ that the user stories are implemented. Then, comment out or remove the extra parts of the tests you wrote and (even though they're good!) try to only write those necessary ones next time. We are happy to provide you feedback on that afterward if you like, just let us know.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+        expect(page).to have_css("div#room-#{@room2.id}")
+        within("div#room-#{@room2.id}") do
+          expect(page).to have_content(@room2.suite)
+          expect(page).to have_content(@room2.rate)
+          expect(page).to have_content(@hotel.name)
+        end
+      end
+
+      @guest2.reload
+      expect(@guest2.rooms).to include(@room2)
```
cjsim89, line 143: This test works great, except doing `.to include(...)` is not really testing that it's on hte page - remember a feature test is looking at the rendered view specifically, not at objects. So, I would rewrite this and line 121 to be
```
expect(page).to have_content(@room2.suite)
```

spec/features/hotels/show_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      expect(page).to have_content(@guest4.name)
+      expect(page).to have_content(@guest5.name)
+      expect(page).to have_content(@guest6.name)
+    end
+
+    visit "/hotels/#{@hotel2.id}"
+
+    within("div#unique_hotel_guests") do
+      expect(page).to have_content(@guest1.name)
+      expect(page).to have_content(@guest2.name)
```
cjsim89, line 60: These tests are again great, but we also want to test that these names only appear in this block _once_ (hence, uniqueness). One way to do this is: 

```
expect(page).to have_content(@guest1.name, count: 1)
```
Or
```
expect(page).to have_content(@guest1.name).once
```

app/models/hotel.rb
ethanrossblack: this is good to know!

spec/features/hotels/show_spec.rb
ethanrossblack: I did not know you could test for this, this is very good to know

spec/features/guests/show_spec.rb
ethanrossblack: I have `expect(page).to have_content(@room2.suite)` on line 136 and sort of on line 123 
_(123 tests for the css div block, but I probably should have tested for content)_

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,5 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
```
cjsim89: Testing

cjsim89:
```suggestion
    @guest = Guest.find(params[:id]) # thing
```

app/models/hotel.rb
cjsim89:
```suggestion
  has_many :guests, through: :rooms
```

PR #12, opened 2023-07-24
Artemy Gibson

algibson1:
I got through the extension user story.
I felt pretty good about this challenge. My main struggle so far in Mod2 is ActiveRecord, and the first three user stories did not involve that. The extension user story did challenge me a little, but I'm excited that I got it to work! I'd love some feedback on the query I used.

I do have a question regarding testing relationships-
both `it {should have_many :guests}` && `it {should have_many(:guests).through(:guest_rooms)}` will pass the test, but the latter further specifies how the relationship works. Is it conventional to use the latter, or is the former considered "enough"?

### DIFF
--- a/app/models/hotel.rb
+++ b/app/models/hotel.rb
@@ -1,3 +1,7 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+
+  def guest_list
+    Guest.distinct.joins(rooms: :hotel).where(hotel: {name: self.name}).pluck(:name).join(", ")
+  end
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -2,4 +2,8 @@ class Room < ApplicationRecord
   belongs_to :hotel
   has_many :guest_rooms
   has_many :guests, through: :guest_rooms
+
+  def count_guests
+    guests.count
+  end
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
   belongs_to :hotel
+  has_many :guest_rooms
+  has_many :guests, through: :guest_rooms
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,4 +3,6 @@ Rails.application.routes.draw do
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/guests/:id", to: "guests#show"
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,4 +5,6 @@ Rails.application.routes.draw do
   # root "articles#index"
 
   get "/guests/:id", to: "guests#show"
+
+  post "/guests/:id", to: "guest_rooms#create"
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -9,4 +9,6 @@ Rails.application.routes.draw do
   post "/guests/:id", to: "guest_rooms#create"
 
   get "/rooms", to: "rooms#index"
+
+  get "/hotels/:id", to: "hotels#show"
 end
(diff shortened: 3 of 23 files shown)


app/models/hotel.rb
```diff
@@ -1,3 +1,7 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+
+  def guest_list
+    Guest.distinct.joins(rooms: :hotel).where(hotel: {name: self.name}).pluck(:name).join(", ")
```
juliet-e, line 5: This is great and correct! Another option would be to start at Hotel and move the other direction:
`Hotel.joins(rooms: :guests).where("rooms.hotel_id = 2").distinct("guests.name").pluck("guests.name")`

However, an even simpler option is to use associations. It's possible in AR to use the `through` associations to chain associations and leverage relationships set up for a resource a model is already associated to. In this case, hotel is associated to rooms, and rooms has associations set up to guests. So, you can add an association to hotel like this: `has_many :guests, through: :rooms`
Then, in your guest_list method, you can simply use `guests.distinct` to grab the guests from a given hotel! Very simple. Very nifty.

app/models/room.rb
```diff
@@ -1,3 +1,9 @@
 class Room < ApplicationRecord
   belongs_to :hotel
+  has_many :guest_rooms
+  has_many :guests, through: :guest_rooms
+
+  def count_guests
+    guests.count
```
juliet-e, line 7: Nice work encapsulating this logic. This could be enhanced with `distinct` so that we don't get duplicate names

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  describe "list of rooms the guest has stayed in" do
+    it "includes room's suite, nightly rate, and name of hotel" do
+      visit "/guests/#{@charlize.id}"
+
+      expect(page).to have_content("This guest has stayed in the following rooms:")
+      expect(page).to_not have_content("Standard")
+      expect(page).to_not have_content("100")
+
+      within("#room-#{@room_1.id}") do
```
juliet-e, line 47: Great use of within blocks! They make this test much more meaningful

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "rails_helper"
+
+RSpec.describe Guest do
+  describe "relationships" do
+    it { should have_many :guest_rooms }
+    it { should have_many(:rooms).through(:guest_rooms) }
```
juliet-e, line 6: Per your comment on this PR: Using this assertion with `through` is more precise of a test. Tables can be associated through different routes (think of a db diagram where there's a circle of relationships so there could be relationships set up from either direction). Typically there's a more direct association that we set up in the model, so we want to test for that precise association path rather than just testing that it's associated.

spec/models/hotel_spec.rb
```diff
@@ -1,7 +1,37 @@
...
+    GuestRoom.create!(guest: @david, room: @room_2)
+    GuestRoom.create!(guest: @david, room: @room_4)
+    GuestRoom.create!(guest: @jim, room: @room_1)
+  end
+
   describe "relationships" do
     it { should have_many :rooms }
   end
+
+  describe "#guest_list" do
```
juliet-e, line 31: Great work with these model tests!

config/routes.rb
```diff
@@ -3,4 +3,12 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/guests/:id", to: "guests#show"
+
+  post "/guests/:id", to: "guest_rooms#create"
```
juliet-e, line 9: A more RESTful route here would be "/guests/:id/rooms" just to indicate that it's posting to a JOINS record rather than creating a new guest (which is what the current path suggests)

PR #13, opened 2023-07-24
Daniel Gallagher

Daniel-Gallagher92:
8.5 / 10 

I feel wayyy better than the first IC, and I'm feeling more confident in knowing where certain things should live. I could have gone for the extension if I hadn't had 20 minutes of debugging my gemfile in the beginning. Thanks Mac OS updates!


app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,17 @@
+class GuestsController < ApplicationController 
+
+  def show 
+    @guest = Guest.find(params[:id])
+    @guest_rooms = Guest.find(params[:id]).rooms.first
+  end
```
megstang, line 6: On line 5, you are choosing the first room off of the collection, but don't we want to show all of the rooms for that guest? 

I think we'll only need to have line 4 in this method. In the view, we can call`@guest.rooms` in order to get the entire collection of rooms that's associated with that guest.

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def create 
+    @guest = Guest.find(params[:id])
+    @new_room = Room.find(params[:room_id])
+
+    @guest.rooms << @new_room
+    @guest.save
+
+    redirect_to "/guests/#{@guest.id}"
+  end
```
megstang, line 16: I like that this logic is in a `create` method, but what kind of resource are we creating? We're actually creating a GuestRoom object (on line 12, you're creating that association, so that's where that resource is getting created). For that reason, it would be most ReSTful to have this in a `GuestRoomsController` in a `create` action. 

Also, we could make this code a bit more optimal. Right now, you're finding the guest and finding the room, and then creating the object. But, you actually have EVERYTHING you need to make that connection without first finding those two objects. You could do the following: 

```ruby
GuestRoom.create(guest_id: params[:id], room_id: params[:room_id])
```

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,8 @@
+class RoomsController < ApplicationController
+  
+  def index
+    @rooms = Room.all
+    @hotel = Room.first.hotel_name
+    @count = Room.first.guest_count
+  end
+end
```
megstang, line 8: Similar to my first comment that I had left above, you are only sending the hotel name and the guest count for the VERY FIRST room in your table. We want to display ALL of the rooms, and each room should display their respective hotel name and guest count. You can imagine, we might have multiple hotels with their own rooms. Right now, the way this is set up, you're only sending one hotels name and one rooms guest count.

So, having line 4 is all you need, since we really just want to send over ALL of our rooms to the view. Then, we'll iterate through that collection and get the respective hotels and guest counts for each room to display.

app/models/guest.rb
```diff
@@ -0,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
```
megstang, line 3: 👍

app/models/guest_room.rb
```diff
@@ -0,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
```

app/models/room.rb
```diff
@@ -1,3 +1,16 @@
 class Room < ApplicationRecord
   belongs_to :hotel
+  
+  has_many :guest_rooms
+  has_many :guests, through: :guest_rooms
```
megstang, line 5: 👍

app/models/room.rb
```diff
@@ -1,3 +1,16 @@
...
+
+  def guest_count
+    guests.count
+  end
+
+  def hotel_name 
+    hotel.name
+  end
```
megstang, line 13: good use of model methods!

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,14 @@
+<h1><%= @guest.name %>'s Stay at <%= @guest_rooms.hotel.name %></h1>
+<% @guest.rooms.each do |room| %>
+  <p>Guest's Room: <%= room.suite %></p>
+  <p>Nightly Rate: <%= room.rate %></p>
+<% end %>
+<p>Number of Nights: <%= @guest.nights %></p>
```
megstang, line 6: Following up on one of my first comments... since we want to display each room's hotel name, we'd want to display that during the each block: 

```html.erb
<h1><%= @guest.name %>'s Stays </h1>
<% @guest.rooms.each do |room| %>
  <p>Guest's Room: <%= room.suite %></p>
  <p>Nightly Rate: <%= room.rate %></p>
  <p>Hotel Name: <%= room.hotel.name %></p>
<% end %>
<p>Number of Nights: <%= @guest.nights %></p>
```

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,10 @@
+<h1> All Rooms at <%= @hotel %></h1> <hr>
+<section class='rooms'>
+<% @rooms.each do |room| %>
+  <p> Suite Name: <%= room.suite %> </p>
+  <p> Nightly Rate: <%= room.rate %> </p>
+  <b> Hotel Name: <%= @hotel %> </b>
+  <p> How many guest's have stayed?: <%= @count %></p><br><hr><hr><br>
+<% end %>
+
+</section>
```
megstang, line 10: So here, we want to use the room object to get that room object's hotel, and count. 

```html.erb
<% @rooms.each do |room| %>
  <p> Suite Name: <%= room.suite %> </p>
  <p> Nightly Rate: <%= room.rate %> </p>
  <b> Hotel Name: <%= room.hotel.name %> </b>
  <p> How many guest's have stayed?: <%= room.guest_count %></p><br><hr><hr><br>
<% end %>
```

config/routes.rb
```diff
@@ -3,4 +3,9 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/guests/:id", to: "guests#show"
+  post "/guests/:id", to: "guests#create"
```
megstang, line 8: Following up on a note that I had written earlier in your `GuestsController`, let's have this route go to a `guest_rooms` controller. Currently, this route is a RESTful route for the creation of a guest resource, so let's leave that route available in case we ever need to add in that functionality. One thing we could do instead is:

```ruby
post '/guests/:id/rooms', to:'guest_rooms#create'
```

spec/models/room_spec.rb
```diff
@@ -1,7 +1,31 @@
...
     it { should belong_to :hotel }
   end
+
+  describe "relationships" do 
+    it { should have_many :guest_rooms }
+  end
+
+  describe "relationships" do
+    it { should have_many(:guests).through(:guest_rooms) }
+  end
```
megstang, line 20: It's fairly common to see all of these `it` blocks on lines 11, 15, and 19 together in one `describe 'relationships'` block. Feel free to do that if you'd like!

PR #14, opened 2023-07-24
Anna Wiley

awiley33:
I got stuck on creating the form for user story 2. I couldn't figure out how to pass the params correctly. I'm not very happy with how I did.

### DIFF
--- a/app/controllers/room_guests_controller.rb
+++ b/app/controllers/room_guests_controller.rb
@@ -1,0 +1,13 @@
+class RoomGuestsController < ApplicationController
+  def create
+    @room_guest = RoomGuest.create!(room_guest_params)
+    redirect_to "/guests/#{@guest.id}"
+  end
+
+
+  private
+
+  def room_guest_params
+    params.permit(:room_id, :guest_id)
+  end
+end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -2,5 +2,15 @@
 <% @guest.rooms.each do |room| %>
 <p><%= room.suite %></p>
 <p><%= room.rate %></p>
-<p><%= room.hotel.name %></p>
+<p><%= room.hotel.name %></p><br><br>
+<% end %>
+
+<br><br>
+
+<h3>Add this guest to a room:</h3>
+<%= form_with url: "/guests/#{@guest.id}/rooms/#{:room_id}", method: :post, params: {room_id: :room_id, guest_id: @guest.id} local: true do |form| %>
+<p><%= form.label "Enter room id here" %></p>
+<p><%= form.text_field :room_id %></p>
+<p><%= form.submit "Submit" %></p>
+
 <% end %>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,6 @@
+<h2><%= @guest.name %></h2>
+<% @guest.rooms.each do |room| %>
+<p><%= room.suite %></p>
+<p><%= room.rate %></p>
+<p><%= room.hotel.name %></p>
+<% end %>
--- a/spec/features/guests/show_spec.rb
+++ b/spec/features/guests/show_spec.rb
@@ -52,8 +52,8 @@ RSpec.describe "Guests Show Page" do
   it "has a form which allows user to add a room to the guest" do
     visit "/guests/#{@guest_5.id}"
     expect(page).to_not have_content(@room_3.suite)
-    
-    fill_in(:room_id, with: @room_3.id)
+    save_and_open_page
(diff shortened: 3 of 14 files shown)


app/controllers/room_guests_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+    redirect_to "/guests/#{@guest.id}"
+  end
+
+
+  private
+
+  def room_guest_params
+    params.permit(:room_id, :guest_id)
+  end
+end
```
megstang, line 13: I like how you are using strong params here. One thing to be careful of, is that a a`!` will render a gross error if for whatever reason a `RoomGuest` isn't able to get created. This could be for multiple reasons.. maybe a user passes in an invalid room id or guest id. Sometimes, I think people think that the bang will force a save, but really, it forces an error IF the object can't be saved to the db.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,16 @@
...
+
+<br><br>
+
+<h3>Add this guest to a room:</h3>
+<%= form_with url: "/guests/#{@guest.id}/rooms/#{:room_id}", method: :post, params: {room_id: :room_id, guest_id: @guest.id} local: true do |form| %>
+<p><%= form.label "Enter room id here" %></p>
+<p><%= form.text_field :room_id %></p>
+<p><%= form.submit "Submit" %></p>
+
+<% end %>
```
megstang, line 16: I know this is where you mentioned you got a bit mixed up. Looking at your routes file, you have `"/guests/:id/rooms/:id"`. These placeholders will get filled and passed along in the `params` hash that you have access to in your controller. But one thing about hashes is that all of your keys will be unique, and if you try to add another key/value pair with the same key as something that exists already in that hash, the value will get overwritten and only one key/value pair will be sent through. I wonder if this is one of the issues you were running into with this implementation. 

That said, I think we could have a route like `POST '/guests/:id/rooms` , and not pass the room_id in the URI because we are depending on the user to pass that id in via the form that you have on this page. If you did it this way, when you got to your controller access and took a look at your params, you'd have the `id` from the URI and you'd have `room_id` from the form field. It seems like maybe one of the main misconceptions you might of had was that you wanted to pass the room id into the URL, but that room_id is provided via the form, not the URI.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    expect(page).to have_content(@guest_3.name)
+    expect(page).to have_content(@room_4.suite)
+  end
+
+  it "displays each room's nightly rate, and the name of the hotel that it belongs to" do
+    visit "/guests/#{@guest_5.id}"
+
+    expect(page).to have_content(@room_1.rate)
+    expect(page).to have_content(@room_2.rate)
+    expect(page).to have_content(@hotel.name).twice
```
megstang, line 44: This is cool! I haven't seen this `twice` method before! thanks!

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+  it "has a form which allows user to add a room to the guest" do
+    visit "/guests/#{@guest_5.id}"
+    expect(page).to_not have_content(@room_3.suite)
+    save_and_open_page
+    fill_in(:room_id, with: "#{@room_3.id}")
+    click_button("Submit")
+    
+    expect(current_path).to eq("/guests/#{@guest_5.id}")
+    expect(page).to have_content(@room_3.suite)
+  end
```
megstang, line 61: Just to follow up once again on the form, we can even see here on line 56, that the `room_id` field will be provided by the user filling out the form, not by the route that they'll be directed to upon hitting submit.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    @room_4 = @hotel.rooms.create!(rate: 250, suite: "Sauna")
+    @room_5 = @hotel.rooms.create!(rate: 250, suite: "Fireplace")
+
+    @room_guest_1 = RoomGuest.create!(room: @room_1, guest: @guest_5)
+    @room_guest_2 = RoomGuest.create!(room: @room_1, guest: @guest_2)
+    @room_guest_3 = RoomGuest.create!(room: @room_2, guest: @guest_5)
+    @room_guest_4 = RoomGuest.create!(room: @room_3, guest: @guest_1)
+    @room_guest_5 = RoomGuest.create!(room: @room_4, guest: @guest_3)
+    @room_guest_6 = RoomGuest.create!(room: @room_5, guest: @guest_4)
+  end
```
megstang, line 25: Great setup, getting multiple rooms on multiple guests. This looks great!

megstang (commented):
@awiley33 I left some comments for you to take a look over. Please let me know if you have any questions!

PR #15, opened 2023-07-24
Allen Russell

garussell:
I took my time setting up the migrations by drawing out the associations and decided to create a guests table, and join table for guests and rooms.  My focus was to follow TDD and MVC convention which I believe I did.  I thought I did a good job using 'pry' and checking what I have access to each step of the way.

The room model has an instance method for 'number_of_guests' that was called in the guests controller 'show', and that controller was called in the guests show page view.

The hotel model utilized a SQL join in the model, which was tested, that joins Guest to guest_rooms in order to access and list guests.  I felt this was necessary because I needed a list of guests, which Hotel does not have access to on its own.  This method was called in the hotels controller, and iterated on the hotel show page view.

I completed the extension and overall felt really good about this IC.  In addition, I am seeing 100% coverage per SimpleCov and did my best to check my work within the allotted time.


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,11 @@
+class GuestRoomsController < ApplicationController
+  def create
+    guest = Guest.find(params[:id])
+    room = Room.find(params[:room_id])
+
+    guest.rooms << room
+    guest.save
```
cjsim89, line 7: Nice job creating this 3rd controller, it's exactly the approach I like to take!

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
```
cjsim89, line 4: This is fine but unnecessary, since we can also access `.rooms` from the view.

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,6 @@
+class HotelsController < ApplicationController
+  def show
+    @hotel = Hotel.find(params[:hotel_id])
+    @guests = @hotel.list_of_guests
```
cjsim89, line 4: Similar to your GuestsController, you can also call the `list_of_guests` method from the view - this is fine though.

app/models/hotel.rb
```diff
@@ -1,3 +1,9 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+
+  def list_of_guests
+    Guest.joins(:guest_rooms)
+    .where(guest_rooms: { room: rooms })
+    .distinct
```
cjsim89, line 7: This query works, nice job! 
One way to make this easier would be to create another association from Hotel to Guests. But how would we do that? 

```
```
^ That works because of the association on `Room` of `has_many :guests, through: :guest_rooms`. Therefore it logically follows that a Hotel, which has many rooms, would also have many guests through those rooms. 

Try adding this in for practice, and adjusting your query to that new association. What would that look like?

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,16 @@
+<%= @guest.name %>
+
+<% @rooms.each do |room| %>
+  <p>Hotel Name: <%= room.hotel.name %></p>
+  <p>Suite: <%= room.suite %></p>
+  <p>Rate: <%= room.rate %></p>
+<% end %>
+  
+<div id="add_guest">
+  <%= form_with url: "/guests/#{@guest.id}", method: :post, local: true do |form| %>
```
cjsim89, line 10: This is the right idea and it works, but I would adjust your path to reflect what exactly you're creating through your `:post` method - something like 
```
"/guests/:guest_id/rooms"
```
would work, because it describes that we're creating a room for this guest by its id. In the end, the actual functionality here is great, but we could improve just the path for this request to be more RESTful and having it describe _what_ it's creating.

cjsim89: Alternatively, your path could just be `"/guest_rooms"`, and either pass in the guest_id through a query param, or through a hidden field in the form.

config/routes.rb
```diff
@@ -3,4 +3,11 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/guests/:id", to: "guests#show"
+  post "/guests/:id", to: "guest_rooms#create"
```
cjsim89, line 8: This would be a tiny MVC issue with the path itself (as I explained above) but overall good work on these routes.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  describe "as a visitor" do
+    describe "when I visit a guest's show page" do
+      it "I see the guest's name, I see a list of all the rooms they've stayed in including the room's suite, nightly rate, and the name of the hotel that it belongs to" do
+        visit "/guests/#{@guest_1.id}"
+
+        expect(page).to have_content(@guest_1.name)
+        expect(page).to have_content(@hotel_1.name)
+        expect(page).to have_content(@room_1.rate)
+        expect(page).to have_content(@room_1.suite)
```
cjsim89, line 23: This test could be strengthened by adding a line checking for which rooms _do not_ belong to this guest, and that they don't appear on the page. Eg: 

```
expect(page).to_not have_content(@room_2.suite)
```
Yours are fine, but examples like this would serve to strengthen the quality of your application by looking at it from different sides.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      end
+    end
+
+    describe "add a guest" do
+      it "I see a form to add a room to this guest. When I fill in a field with the id of an existing room, And I click submit, then I am redirected back to the guest's show page and see the room listed uder this guest's rooms" do
+        visit "/guests/#{@guest_1.id}"
+
+        expect(page).to_not have_content(@room_2.suite)
```
cjsim89, line 31: Nice job adding this line!

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        within("#add_guest") do
```
cjsim89, line 32: Good use of a `within` block

spec/features/hotels/show_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    GuestRoom.create!(guest: @guest_1, room: @room_1)
+    GuestRoom.create!(guest: @guest_2, room: @room_1)
+  end
+
+  describe "as a visitor" do
+    describe "when I visit a hotel's show page"
+      it "I see a unique list of all guests that have stayed at this hotel" do
+        visit "/hotels/#{@hotel_1.id}" 
+
+        expect(page).to have_content(@guest_1.name)
```
cjsim89, line 22: Nice job with the extension; the only thing missing is checking for _uniqueness_ in this feature test. One way to do it would be to add `count: 1` to the end of the `have_content` method as a 2nd argument. Ex: 

```
```

This way, if the # of times this content appears on the page is _not exactly 1_, the test will fail.

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    @room_1 = @hotel_1.rooms.create!(rate: 125, suite: "Economy")
+    @room_2 = @hotel_1.rooms.create!(rate: 500, suite: "Presidential")
+    
+    GuestRoom.create!(guest: @guest_1, room: @room_1)
+    GuestRoom.create!(guest: @guest_2, room: @room_1)
+  end
+
+  describe "as a visitor" do
+    describe "when I visit the rooms index page" do
+      it "I see a list of all rooms including the room's suite, nightly rate, and the name of the hotel that it belongs to and the number of guests that have stayed in that room" do
```
cjsim89, line 19: This test is fine but only proves one room is on the page; you should find a way to test for the 2nd room objects' data as well.

spec/models/hotel_spec.rb
```diff
@@ -1,7 +1,28 @@
...
+    GuestRoom.create!(guest: @guest_2, room: @room_1)
+  end
+
   describe "relationships" do
     it { should have_many :rooms }
   end
+
+  describe "#instance_methods" do
+    it "can list all guests of the hotel" do
+      expect(@hotel_1.list_of_guests).to eq([@guest_1, @guest_2])
```
cjsim89, line 25: This test is fine, thanks for remembering to test your model methods! 

One way to make this better would be to have a 2nd hotel with no rooms attached to it, then running a test against that 2nd instance, returning an empty array. It's little tests like these that can really help strengthen the quality of your tests, therefore making the way you think about programming and the construction of your applications that much better.

PR #16, opened 2023-07-24
Connor Richmond

ConnorRichmond:
I think I am getting better at these though I realize I stuck too much on just trying to make the challenge work and not holding to MVC as I should.

I got through User Story 3

### DIFF
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,0 +1,5 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+  end
+end
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -2,4 +2,14 @@ class GuestsController < ApplicationController
   def show
     @guest = Guest.find(params[:id])
   end
+
+  def add_room
+    @guest = Guest.find(params[:id])
+    room_id = params[:room_id]
+
+    room = Room.find_by(id: room_id)
+    @guest.rooms << room if room
+
+    redirect_to "/guests/#{params[:id]}"
+  end
 end
--- a/app/models/guests_room.rb
+++ b/app/models/guests_room.rb
@@ -1,0 +1,4 @@
+class GuestsRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
--- a/app/views/rooms/index.html.erb
+++ b/app/views/rooms/index.html.erb
@@ -1,0 +1,12 @@
+<h1>Rooms Index</h1>
+
+<ul>
+  <% @rooms.each do |room| %>
+    <li>
+      <strong>Suite:</strong> <%= room.suite %><br>
+      <strong>Nightly Rate:</strong> $<%= room.rate %><br>
+      <strong>Hotel:</strong> <%= room.hotel.name %><br>
+      <strong>Number of Guests:</strong> <%= room.guests.count %>
+    </li>
+  <% end %>
+</ul>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,4 +6,6 @@ Rails.application.routes.draw do
 
   get "/guests/:id", to: "guests#show"
   post "/guests/:id/add_rooms", to: "guests#add_room"
+
+  get "/rooms", to: "rooms#index"
 end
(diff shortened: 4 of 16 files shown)


app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,15 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+  end
+
+  def add_room
```
juliet-e, line 6: A more RESTful option here would be to have a GuestRoom controller with a `create` action. By having a create or even the `add_room` action in the guests controller like you have here, it suggests you're adding a Guest resource, but you're adding a joins record. So having a joins controller with its own create action makes it very clear to other developers what the method will be doing

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+    @guest = Guest.find(params[:id])
+    room_id = params[:room_id]
+
+    room = Room.find_by(id: room_id)
+    @guest.rooms << room if room
```
juliet-e, line 11: using `if room` is a nice way to verify the validity of the ID - not required but nice!

app/models/guests_room.rb
```diff
@@ -0,0 +1,4 @@
+class GuestsRoom < ApplicationRecord
```
juliet-e, line 1: We typically name join models with singular - GuestRoom - and the corresponding table has a plural but only for the second resource - guest_rooms.

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1>Rooms Index</h1>
+
+<ul>
+  <% @rooms.each do |room| %>
+    <li>
+      <strong>Suite:</strong> <%= room.suite %><br>
+      <strong>Nightly Rate:</strong> $<%= room.rate %><br>
+      <strong>Hotel:</strong> <%= room.hotel.name %><br>
+      <strong>Number of Guests:</strong> <%= room.guests.count %>
```
juliet-e, line 9: While it's a simple calculation, having `count` here is still a calculation that deals with querying the database so we'd rather you encapsulate it into a model method. This also allows you to test it in a more isolated environment (model test) which can be very helpful if this were to get more complicated, i.e. if you had to only count distinct guests or there were other conditions on it. I'd consider this a minor MVC infraction

config/routes.rb
```diff
@@ -3,4 +3,9 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/guests/:id", to: "guests#show"
+  post "/guests/:id/add_rooms", to: "guests#add_room"
```
juliet-e, line 8: See note in controller about RESTful-ness

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      @guests_room = GuestsRoom.create(guest: @guest1, room: @room1)
+    end
+
+    it "shows guests with information on show page" do
+      visit "/guests/#{@guest1.id}"
+
+      expect(page).to have_content(@guest1.name)
+      expect(page).to have_content("Hotel: #{@hotel.name}")
+      expect(page).to have_content("Suite: #{@room1.suite}")
+      expect(page).to have_content("Nightly Rate: $#{@room1.rate}")
```
juliet-e, line 21: Ideally when we're testing anything with iteration (in this case, rooms), we'd like to see multiple items displayed in the test. So it would be good to have another GuestRoom record for the 2nd room and then test that room info for both rooms is present. `within` blocks could help you verify that info is in the correct place, as well.

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    GuestsRoom.create(guest: @guest1, room: @room1)
+    GuestsRoom.create(guest: @guest2, room: @room2)
+  end
+
+  it "displays a list of all rooms with suite, nightly rate, hotel name, and number of guests" do
+    visit "/rooms"
+
+    expect(page).to have_content(@room1.suite)
+    expect(page).to have_content(@room1.rate)
```
juliet-e, line 21: These assertions would be more robust with `within` blocks and/or with assertions that include the fixed text around the dynamic values, i.e. `expect(page).to have_content("Nightly Rate: #{@room1.rate}")`
This helps us verify that the data is not only there, but also located in the right place rather than floating randomly on the page

PR #17, opened 2023-07-24
Weston Schutt

westonio:
I started this challenge feeling as though I would not be able to complete the 3 user stories in time (as setup felt like it took forever). However, I surprised myself by completing all user stories (including the extension) and I even took some time to create the _add room to guest show page_  sad path

### DIFF
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,0 +1,7 @@
+class GuestRoomsController <ApplicationController
+  def create
+    GuestRoom.create(guest_id: params[:guest_id], room_id: params[:room_id])
+
+    redirect_to "/guests/#{params[:guest_id]}"
+  end
+end
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,7 +1,13 @@
 class GuestRoomsController <ApplicationController
   def create
-    GuestRoom.create(guest_id: params[:guest_id], room_id: params[:room_id])
-
-    redirect_to "/guests/#{params[:guest_id]}"
+    begin 
+      guest_room = GuestRoom.new(guest_id: params[:guest_id], room_id: params[:room_id])
+      guest_room.save!
+      redirect_to "/guests/#{params[:guest_id]}"
+      flash[:notice] = "Room added successfully!"
+    rescue => exception
+      redirect_to "/guests/#{params[:guest_id]}"
+      flash[:alert] = "Room ID invalid - Please enter existing room ID"
+    end
   end
-end
+end 
--- a/app/models/hotel.rb
+++ b/app/models/hotel.rb
@@ -1,3 +1,9 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+
+  def prev_guests
+    Guest.joins(rooms: :hotel)
+    .distinct
+    .pluck(:name)
+  end
 end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,10 +1,18 @@
 <h1>Welcome back, <%= @guest.name %></h1>
 
-<div id="past_stays">
-  <h2>Previous Room Stays:</h2>
+<div id="rooms">
+  <h2>Booked Rooms:</h2>
   <% @guest.rooms.each do |room| %>
     <h3><%= room.suite %></h3>
     <p>Nightly Rate: $<%= room.rate %></p>
     <p>Room at: <%= room.hotel.name %></p>
   <% end %>
 </div>
+
+<%= form_with url: "/guest_rooms", id: "add_room_form", method: :post, local: :true do |form| %>
+  <%= form.label :room_id, "Enter a Room ID:" %>
+  <%= form.number_field :room_id %>
+  <%= form.hidden_field :guest_id, value: @guest.id %>
+
+  <%= form.submit "Add Room to #{@guest.name}" %>
+<% end %>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,10 @@
+<h1>Welcome back, <%= @guest.name %></h1>
+
+<div id="past_stays">
+  <h2>Previous Room Stays:</h2>
+  <% @guest.rooms.each do |room| %>
+    <h3><%= room.suite %></h3>
+    <p>Nightly Rate: $<%= room.rate %></p>
+    <p>Room at: <%= room.hotel.name %></p>
+  <% end %>
+</div>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -11,5 +11,12 @@
 
   <body>
     <%= yield %>
+    <% if flash.any? %>
+      <% flash.each do |type, msg| %>
+        <div>
+        <%= msg %>
+        </div>
+      <% end %>
+    <% end %>
   </body>
 </html>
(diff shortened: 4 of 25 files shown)


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,13 @@
+class GuestRoomsController <ApplicationController
+  def create
+    begin 
+      guest_room = GuestRoom.new(guest_id: params[:guest_id], room_id: params[:room_id])
+      guest_room.save!
+      redirect_to "/guests/#{params[:guest_id]}"
+      flash[:notice] = "Room added successfully!"
```
cjsim89, line 7: Gotta say, this is the first time I'm seeing `rescue` and error handling being used on an IC in mod 2! This looks good, and we won't really be using this sort of thing until later on in mod 3. 
Overall this is fine, and I see you're implementing the rescue because of the `save!` on line 5. To make your flash message a bit more specific, you can utilize the `errors` method on the `guest_room` instance instead. This would work even better with validations on that model (which we didn't ask you for, so it's ok).

app/models/hotel.rb
```diff
@@ -1,3 +1,9 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+
+  def prev_guests
+    Guest.joins(rooms: :hotel)
```
cjsim89, line 5: This is very close, but since we're doing `Guest` as a class and not an instance, nor checking if the guests have a relationship to _this specific Hotel instance_, then we're seeing ALL guests regardless of where they've stayed, which is not quite what the user story is asking for (` I see a list of all guests that have stayed at this hotel`). 

Instead, you could add a `.where("rooms.hotel_id = #{self.id}")` clause to this query. Or, you could use my preferred method, which is to add second association to this model like so: 
```
```

This method works because a Room already `has_many :guests`, so it logically follows that a Hotel also has many guests through its rooms. This is where my "follow the treasure map" analogy comes in handy.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,18 @@
...
+    <h3><%= room.suite %></h3>
+    <p>Nightly Rate: $<%= room.rate %></p>
+    <p>Room at: <%= room.hotel.name %></p>
+  <% end %>
+</div>
+
+<%= form_with url: "/guest_rooms", id: "add_room_form", method: :post, local: :true do |form| %>
+  <%= form.label :room_id, "Enter a Room ID:" %>
+  <%= form.number_field :room_id %>
+  <%= form.hidden_field :guest_id, value: @guest.id %>
```
cjsim89, line 15: Great job with this form setup!

app/views/layouts/application.html.erb
```diff
@@ -11,5 +11,12 @@
 
   <body>
     <%= yield %>
+    <% if flash.any? %>
```
cjsim89, line 14: This is fine but not required; `.each` will iterate through any size of array, and if it's empty, it just won't iterate.

config/routes.rb
```diff
@@ -3,4 +3,9 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+
+  get "/guests/:id", to: "guests#show"
+  post "/guest_rooms", to: "guest_rooms#create"
+  get "/rooms", to: "rooms#index"
+  get "/hotels/:id", to: "hotels#show"
```
cjsim89, line 10: Good restful routes!

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+      @guest_2 = Guest.create!(name: "Latrice Royal", nights: 2)
+      @guest_3 = Guest.create!(name: "Monet Exchange", nights: 5)
+
+      GuestRoom.create!(guest_id: @guest_1.id, room_id: @room_4.id)
+      GuestRoom.create!(guest_id: @guest_1.id, room_id: @room_1.id)
+      GuestRoom.create!(guest_id: @guest_1.id, room_id: @room_7.id)
+      GuestRoom.create!(guest_id: @guest_2.id, room_id: @room_3.id)
+      GuestRoom.create!(guest_id: @guest_2.id, room_id: @room_6.id)
+      GuestRoom.create!(guest_id: @guest_2.id, room_id: @room_2.id)
+      GuestRoom.create!(guest_id: @guest_2.id, room_id: @room_5.id)
```
cjsim89, line 28: Great job with this object setup! Keep this up. (You may even want to consider downsizing the amount of rooms created to 3-4 instead, to save time)

cjsim89: Or use a gem like Faker to help with values

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+      expect(page).to have_content(@guest_1.name)
+      expect(page).to_not have_content(@guest_2.name)
+    end
+
+    describe "lists all the rooms they've stayed in" do
+      it "lists the room's suite" do
+        within("#rooms") do
+          expect(page).to have_content(@room_4.suite)
+          expect(page).to have_content(@room_1.suite)
+          expect(page).to have_content(@room_7.suite)
```
cjsim89, line 43: Try to throw in a `to_not` to use all the objects you have in your setup.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+        within("#add_room_form") do
+          fill_in :room_id, with: 6
+          click_button("Add Room to #{@guest_1.name}")
+        end
+        
+        expect(current_path).to eq("/guests/#{@guest_1.id}")
+      end
+
+      it 'shows the new room now listed under the guests past stays' do
+        expect(page).to_not have_content(@room_6.suite)
```
cjsim89, line 80: Excellent use of `to_not` here to prove the opposite that you want to prove later on!

spec/features/hotels/show_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      GuestRoom.create!(guest_id: @guest_2.id, room_id: @room_2.id)
+      GuestRoom.create!(guest_id: @guest_4.id, room_id: @room_2.id)
+      GuestRoom.create!(guest_id: @guest_3.id, room_id: @room_2.id)
+
+      visit "/hotels/#{@beverly.id}"
+    end
+    
+    it "shows a unique list of all guests that have stayed at this hotel" do
+      within("#guest-list") do
+        expect(page).to have_content(@guest_1.name)
```
cjsim89, line 34: This is a good test, but to really test for uniqueness with capybara, you'll want to include `,count: 1` as a 2nd argument to the `have_content` method. Ex: 

```
```

This way it should only find one instance of the first argument on the page - if there's more (or less), the test will fail.

spec/models/hotel_spec.rb
```diff
@@ -4,4 +4,37 @@
...
+      GuestRoom.create!(guest_id: @guest_2.id, room_id: @room_2.id)
+      GuestRoom.create!(guest_id: @guest_2.id, room_id: @room_2.id)
+      GuestRoom.create!(guest_id: @guest_4.id, room_id: @room_2.id)
+      GuestRoom.create!(guest_id: @guest_3.id, room_id: @room_2.id)
+    end
+
+    it 'can create a unique list of all previous guests who have stayed within the hotels rooms' do
+      expected = [@guest_5.name, @guest_4.name, @guest_2.name, @guest_1.name, @guest_3.name]
+
+      expect(@beverly.prev_guests.sort).to eq(expected)
```
cjsim89, line 37: This test passes because you only have 1 Hotel, but if you had a 2nd hotel, it would not pass because of your use of the `Guest` class (essentially doing `Guest.all`) in your method. Try it out and let me know what you think.

app/models/hotel.rb
westonio: This makes so much sense, and I see how I could have written my tests better to catch this. Thank you!

PR #18, opened 2023-07-24
Derek Chavez

DChavez18:
I made it through the 3 user stories

The part i struggled with the most was setting up the form to add a room to a guest via id,
otherwise I felt pretty good about the IC

### DIFF
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -3,4 +3,6 @@ class GuestsController < ApplicationController
     @guest = Guest.find(params[:id])
     @rooms = @guest.rooms
   end
+  
+
 end
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,0 +1,6 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
+  end
+end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -10,3 +10,11 @@
         <p>Hotel Name:</p> <%= room.hotel.name %>
     </li>
   <% end %>
+
+<h2>Add a Room</h2>
+  <%= form_with url: "/guests/#{@guest.id}", method: :post, local: true do |f| %>
+    <%= f.label :room_name, "Room Name" %>
+    <%= f.text_field :room_name %>
+
+    <%= f.submit "Add Room" %>
+  <% end %>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,12 @@
+<h1>Guest Info</h1>
+<h2> Guest Name: <%= @guest.name %></h2>
+
+<h2>Rooms Enjoyed</h2>
+
+  <% @guest.rooms.each do |room| %>
+    <li>
+        <p>Room Suite:</p> <%= room.suite %><br>
+        <p>Nightly Rate:</p> <%= room.rate %><br>
+        <p>Hotel Name:</p> <%= room.hotel.name %>
+    </li>
+  <% end %>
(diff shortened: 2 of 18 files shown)


app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,8 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
```
juliet-e, line 4: Keep in mind that you don't necessarily need to set up the @rooms instance var here as you can reference it in the view as @guest.rooms

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,20 @@
+<h1>Guest Info</h1>
+<h2> Guest Name: <%= @guest.name %></h2>
+
+<h2>Rooms Enjoyed</h2>
+
+  <% @guest.rooms.each do |room| %>
+    <li>
+        <p>Room Suite:</p> <%= room.suite %><br>
+        <p>Nightly Rate:</p> <%= room.rate %><br>
+        <p>Hotel Name:</p> <%= room.hotel.name %>
```
juliet-e, line 10: Nice work reaching through objects to get the hotel name!

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1>All Roons</h1>
+
+<ul>
+  <% @rooms.each do |room| %>
+    <li>
+      <p>Suite: <%= room.suite %></p>
+      <p>Nightly Rate: <%= room.rate %></p>
+      <p>Hotel Name: <%= room.hotel.name %></p>
+      <p> Guest Count: <%= room.guest_count %></p>
```
juliet-e, line 9: Great work encapsulating the counting logic in the model.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+# Guest Show
+# As a visitor
+# When I visit a guest's show page
+# I see the guest's name
+# And I see a list of all the rooms they've stayed in
+# including the room's suite, nightly rate, 
+# and the name of the hotel that it belongs to.
+
+  it "displays the guest's name" do
+    hotel_1 = Hotel.create!(name: "Aspen Inn", location: "Aspen" )
```
juliet-e, line 15: I'd recommend using a before each so that you don't have to copy and paste the test set up in each test case

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    guest_1 = Guest.create!(name: "Charlie Day", nights: 3)
+
+    guest_1.rooms << room_1
+    guest_1.rooms << room_2
+
+    visit "/guests/#{guest_1.id}"
+
+
+    expect(page).to have_content("Presidential")
```
juliet-e, line 38: These tests could be more robust if you include the fixed text that's on the same line, so that you can be sure the dynamic parts of the view are located in the right place. For example:
`expect(page).to have_content("Room Suite: Presidential")`

juliet-e: within blocks could be used here too to indicate which room it's referring to

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    room_1.guests << guest_1
+    room_2.guests << guest_1
+    room_2.guests << guest_2
+    room_3.guests << guest_1
+    room_3.guests << guest_2
+    room_3.guests << guest_3
+
+    visit "/rooms"
+
+    expect(page).to have_content("Guest Count: 1")
```
juliet-e, line 64: Within blocks would be specifically helpful here, as without them it's unclear which room these counts are referring to. When we have this model tested, we can be sure the method is working, but it's still a good practice to differentiate parts of a page for this type of assertion

spec/models/room_spec.rb
```diff
@@ -3,5 +3,23 @@
 RSpec.describe Room, type: :model do
   describe "relationships" do
     it { should belong_to :hotel }
+    it { should have_many :guest_rooms }
+    it { should have_many(:guests).through(:guest_rooms)}
+  end
+
+  describe "#guest_count" do
+    it "counts all of the guests that belong to a room" do
```
juliet-e, line 11: Great work remembering to test the model methods!

PR #19, opened 2023-07-24
andi lovetto

andilovetto:
completes user story 3; had difficulty adding params to form_with url. tried several methods for adding the guest_id into the params; i eventually added it to the route, which is likely not correct. got the job done either way, lol! please advise the restful approach for accomplishing this. thanks!

### DIFF
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,0 +1,9 @@
+class GuestRoomsController < ApplicationController
+
+  def create
+   GuestRoom.create!(guest_id: params[:id], room_id: params[:room_id])
+   
+   redirect_to "/guests/#{params[:id]}"
+  end
+
+end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -1,3 +1,6 @@
 class Room < ApplicationRecord
   belongs_to :hotel
+
+  has_many :guest_rooms
+  has_many :guests, through: :guest_rooms
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -3,4 +3,11 @@ class Room < ApplicationRecord
 
   has_many :guest_rooms
   has_many :guests, through: :guest_rooms
+
+
+  def total_guests
+    guests.count
+  end
+
 end
+
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,6 @@
+<h1>Guest Name: <%= @guest.name %></h1>
+<% @guest.rooms.each do |room| %>
+  <p>Suite: <%= room.suite %></p>
+  <p>Nightly Rate: <%= room.rate %></p>
+  <p>Hotel: <%= room.hotel.name %></p>
+<% end %>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -4,3 +4,10 @@
   <p>Nightly Rate: <%= room.rate %></p>
   <p>Hotel: <%= room.hotel.name %></p>
 <% end %>
+<div class="new_room">
+  <%= form_with url: "/guest_rooms/#{@guest.id}", method: :post, local: true do |f| %>
+    <%= f.label :room_id, "Room id" %>
+    <%= f.text_field :room_id %>
+    <%= f.submit %>
+  <% end %>
+</div>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,4 +3,6 @@ Rails.application.routes.draw do
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "/guests/:id", to: "guests#show"
+
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,4 +5,6 @@ Rails.application.routes.draw do
   # root "articles#index"
   get "/guests/:id", to: "guests#show"
 
+  post "/guest_rooms/:id", to: "guest_rooms#create"
+
 end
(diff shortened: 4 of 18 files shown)


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,9 @@
+class GuestRoomsController < ApplicationController
+
+  def create
+   GuestRoom.create!(guest_id: params[:id], room_id: params[:room_id])
```
juliet-e, line 4: This is totally fine for an IC, but in projects and in the future, try to use `create!` only in tests or seeds files where we don't care if exceptions are thrown if the data passed in is wrong. In controllers, we want to make sure users have a more friendly experience so we want just `create` (or on a project you might do some error handling with flash messages)

app/models/room.rb
```diff
@@ -1,3 +1,13 @@
...
   belongs_to :hotel
-end
\ No newline at end of file
+
+  has_many :guest_rooms
+  has_many :guests, through: :guest_rooms
+
+
+  def total_guests
+    guests.count
```
juliet-e, line 9: Great work encapsulating this logic in a model method!

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,13 @@
+<h1>Guest Name: <%= @guest.name %></h1>
+<% @guest.rooms.each do |room| %>
+  <p>Suite: <%= room.suite %></p>
+  <p>Nightly Rate: <%= room.rate %></p>
+  <p>Hotel: <%= room.hotel.name %></p>
+<% end %>
+<div class="new_room">
+  <%= form_with url: "/guest_rooms/#{@guest.id}", method: :post, local: true do |f| %>
```
juliet-e, line 8: To get the guest id passed in the request, you could add it to the path but a more RESTful route would be `"/guests/guest_id/rooms"` to indicate that the id in the path is for the guest, not the guest_rooms. Another option is using [hidden](https://stackoverflow.com/questions/64533667/how-to-add-a-hidden-field-to-form-with) fields in erb forms

config/routes.rb
```diff
@@ -3,4 +3,10 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "/guests/:id", to: "guests#show"
+
+  post "/guest_rooms/:id", to: "guest_rooms#create"
```
juliet-e, line 8: see note about path in your view for the guest show page

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,39 @@
+require "rails_helper"
+
+RSpec.describe "the guests show page" do
+  let!(:guest_1) { Guest.create!(name: "Andi", nights: 2) }
```
juliet-e, line 4: Oh nice use of `let` initializing!

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  let!(:guest_room_1) { GuestRoom.create!(guest_id: guest_1.id, room_id: room_1.id) }
+  let!(:guest_room_2) { GuestRoom.create!(guest_id: guest_1.id, room_id: room_2.id) }
+
+  before do
+    visit "/guests/#{guest_1.id}"
+  end
+
+
+  it "shows guest name and list of rooms and room attributes with hotel name" do
+    expect(page).to have_content("Guest Name: #{guest_1.name}")
```
juliet-e, line 18: within blocks could be helpful here to break up these assertions and ensure they're grouped together properly

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    expect(page).to have_content("Suite: #{room_2.suite}")
+    expect(page).to have_content("Nightly Rate: #{room_2.rate}")
+    expect(page).to have_content("Hotel: #{hotel_1.name}")
+    expect(page).to_not have_content("Suite: #{room_3.suite}")
+    expect(page).to_not have_content("Nightly Rate: #{room_3.rate}")
+  end
+
+  it "has a form for adding a room to a guest" do
+    within(".new_room") do
+      fill_in :id, with: room_3.id
```
juliet-e, line 30: Nice use of `within` here!

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      expect(page).to have_content(hotel_1.name)
+      expect(page).to have_content("Total Guests: 1")
+    end
+
+    within(".room-#{room_3.id}") do
+      expect(page).to have_content(room_3.suite)
+      expect(page).to have_content(room_3.rate)
+      expect(page).to have_content(hotel_1.name)
+      expect(page).to have_content("Total Guests: 1")
+    end
```
juliet-e, line 40: Great use of within blocks here

spec/models/room_spec.rb
```diff
@@ -1,7 +1,32 @@
...
+    let!(:guest_room_1) { GuestRoom.create!(guest_id: guest_1.id, room_id: room_1.id) }
+    let!(:guest_room_2) { GuestRoom.create!(guest_id: guest_1.id, room_id: room_2.id) }
+    let!(:guest_room_3) { GuestRoom.create!(guest_id: guest_2.id, room_id: room_1.id) }
+    let!(:guest_room_4) { GuestRoom.create!(guest_id: guest_2.id, room_id: room_3.id) }
+
+    describe ".total_guests" do
+      it "returns the number of guests" do
+        expect(room_1.total_guests).to eq(2)
+        expect(room_2.total_guests).to eq(1)
+        expect(room_3.total_guests).to eq(1)
```
juliet-e, line 28: Great work setting up meaningful test data for this model method!

PR #20, opened 2023-07-24
Mike Wood

MWoodshop:
I got through User Story 3. Instead of moving on to the extension I refactored the code and found several places I could improve, add a test, and fix a potential MVC violation. Overall I feel confident this code fulfills the requirements of the user stories.


app/models/guest.rb
```diff
@@ -0,0 +1,8 @@
+class Guest < ApplicationRecord
+  has_many :stays
+  has_many :rooms, through: :stays
+
+  def add_room(room_id)
+    rooms << Room.find(room_id)
```
cjsim89, line 6: This operation can be done in a controller. If you are making a model method, you'll need to test it. And when you go to test this method, you'll find there's really nothing that this method is doing that Rails doesn't already do under-the-hood - it's creating a new relationship (a new `Stay` instance as you called it). 

Also, I don't see any model methods being tested or model tests for your relationships...! This is a habit I greatly encourage you to break. Try to stick to TDD for all models and all lines of each models as you go.

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  def index
+    @guests = Guest.all
+  end
+
+  def show
+    @guest = Guest.find(params[:id])
+    @stays = @guest.stays
+  end
+
+  def add_room
```
cjsim89, line 11: The content of this method is great, however an action called `add_room` is not considered RESTful - remember the RESTful actions are `index, show, new, create, edit, update, delete`. If you find yourself wanting to create an action outside of these 7, chances are you'll want to put it in a different controller. 

To use this scenario as an example, the code in this action could be entirely lifted and placed in a `create` action in the `stays_controller`, since that's what we're doing - creating a new Stay instance with a guest ID and a room ID.

app/models/room.rb
```diff
@@ -1,3 +1,9 @@
 class Room < ApplicationRecord
   belongs_to :hotel
-end
\ No newline at end of file
+  has_many :stays
+  has_many :guests, through: :stays
+
+  def distinct_guest_count
+    stays.distinct.count(:guest_id)
```
cjsim89, line 7: Good AR here!

app/models/room.rb
```diff
@@ -1,3 +1,9 @@
 class Room < ApplicationRecord
   belongs_to :hotel
-end
\ No newline at end of file
+  has_many :stays
+  has_many :guests, through: :stays
```
cjsim89, line 4: Nice job creating this association, makes things a lot easier!

config/routes.rb
```diff
@@ -3,4 +3,32 @@
...
+  get '/rooms/:id', to: 'rooms#show', as: 'room'
+  get '/rooms/:id/edit', to: 'rooms#edit', as: 'edit_room'
+  patch '/rooms/update', to: 'rooms#update', as: 'update_room'
+  delete '/rooms/delete', to: 'rooms#destroy', as: 'delete_room'
+
+  # Guest
+  get '/guests', to: 'guests#index', as: 'guests'
+  get '/guests/:id', to: 'guests#show', as: 'guest'
+  get '/guests/:id/edit', to: 'guests#edit', as: 'edit_guest'
+  patch '/guests/:id', to: 'guests#update', as: 'update_guest'
```
cjsim89, line 27: I think I left a comment similar to this before - it is _**not**_ good practice to fill out the routes file with every single  possible route; it leaves your application open to bad actors. From now on, please only fill out this file with the routes that are necessary to fulfill the needs of the user stories. 
This morning you'll learn more about this, and if you want to learn more about it, please ask during that class, and read more in [this article](https://guides.rubyonrails.org/routing.html#restricting-the-routes-created).

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    it 'can add a room to a guest' do
+      visit guest_path(@guest1)
+
+      expect(page).to have_content('Add a Room to this Guest')
+      fill_in 'room_id', with: @room2.id
+      click_button 'Add Room to Guest'
+
+      expect(current_path).to eq(guest_path(@guest1))
+      expect(page).to have_content(@room2.suite)
```
cjsim89, line 38: One way to improve this test would be to check that `@room2.suite` does _not_ appear on the page before you `click_button`. That way it gives more meaning to this line.

spec/features/rooms/show_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    @guest1 = Guest.create(name: 'John Doe', nights: 3)
+    @guest2 = Guest.create(name: 'Jane Doe', nights: 2)
+
+    @stay1 = Stay.create(guest: @guest1, room: @room1)
+    @stay2 = Stay.create(guest: @guest2, room: @room2)
+  end
+
+  # User Story 3 Test
+  describe 'rooms show page' do
```
cjsim89, line 19: I think you may have misread the user story; it asks for this information on the Rooms Index page, not the rooms show. This is partially why I recommend copying and pasting the user story into your feature test files.

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    @guest1 = Guest.create(name: 'John Doe', nights: 3)
+    @guest2 = Guest.create(name: 'Jane Doe', nights: 2)
+
+    @stay1 = Stay.create(guest: @guest1, room: @room1)
+    @stay2 = Stay.create(guest: @guest2, room: @room2)
+  end
+
+  describe 'rooms index page' do
+    it 'can see all room data including distinct guest count' do
```
cjsim89, line 19: This doesn't cover the scope of the user story, which reads: 
```
When I visit the rooms index page
Then I see a list of all rooms
including the room's suite, nightly rate, and the name of the hotel that it belongs to
and the number of guests that have stayed in that room.
```

From this test, you're missing the 2nd room instance and its attributes, including the distinct guest count. (Though, I do see it's in the test marked `show` page. 

I don't say all of this to be a pedant, I'm ok with splitting up and organizing your tests in whatever way suits you best. However, potential employers are going to want to see that you have a sharp attention to detail and can meet their and their clients' expectations with precision. As this and other tests read, I'm not sure I'm seeing that at this point. This is why I am suggesting you copy and paste your user stories in the test file themselves, and go line-by-line in those stories constructing your tests. You can even leave the pasted story in the test for reference.

config/routes.rb
```diff
@@ -3,4 +3,32 @@
...
+  post '/guests/:id/add_room', to: 'guests#add_room', as: 'add_room_guest'
```
cjsim89, line 28: A better route might be `post "/guests/:id/rooms`, which would go to the `stays#create` controller & action.

PR #22, opened 2023-07-24
Wil Fady midmod IC

fadwil:
I spent too long getting the migrations and relationships setup. I could only make it through story 2. If I hadn't spend so much time in the beginning on setup I think I could have completed story 3 and refactored my code as well as tests to be more robust

### DIFF
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,0 +1,6 @@
+class GuestRoomsController < ApplicationController
+  def create
+    guest = GuestRoom.create!(guest_id: params[:id], room_id: params[:room_id])
+    redirect_to "/guests/#{params[:id]}"
+  end
+end
--- a/app/models/guest.rb
+++ b/app/models/guest.rb
@@ -1,0 +1,3 @@
+class Guest < ApplicationRecord
+  has_many :rooms
+end
--- a/app/models/guest.rb
+++ b/app/models/guest.rb
@@ -1,3 +1,4 @@
 class Guest < ApplicationRecord
-  has_many :rooms
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
 end
--- a/app/models/guest_room.rb
+++ b/app/models/guest_room.rb
@@ -1,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -1,4 +1,5 @@
 class Room < ApplicationRecord
   belongs_to :hotel
-  has_many :guests
+  has_many :guest_rooms
+  has_many :guests, through: :guest_rooms
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -1,3 +1,4 @@
 class Room < ApplicationRecord
   belongs_to :hotel
+  has_many :guests
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -4,4 +4,5 @@ Rails.application.routes.draw do
   # Defines the root path route ("/")
   # root "articles#index"
   get "guests/:id", to: "guests#show"
+  post "guests/:id", to: "guest_rooms#create"
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,4 +3,5 @@ Rails.application.routes.draw do
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "guests/:id", to: "guests#show"
 end
--- a/spec/features/guests/show_spec.rb
+++ b/spec/features/guests/show_spec.rb
@@ -1,15 +1,29 @@
 require "rails_helper"
 
-RSpec.describe "/guest/:id", type: :feature do
+RSpec.describe "/guests/:id", type: :feature do
   before do
     @hotel_1 = Hotel.create!(name: 'Echo Mountain Inn', location: 'Echo Mountain')
     @hotel_2 = Hotel.create!(name: 'Hot Springs Inn', location: 'Hot Springs')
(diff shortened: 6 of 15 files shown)


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestRoomsController < ApplicationController
+  def create
+    guest = GuestRoom.create!(guest_id: params[:id], room_id: params[:room_id])
```
megstang, line 3: doesn't look like you're using this `guest` variable in this method, so feel free to get rid of it!

app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,6 @@
...
+    redirect_to "/guests/#{params[:id]}"
+  end
```
megstang, line 5: One thing to be careful of with the ! on line 3 is that your code will throw a gross user-side error (on the browser) if for whatever reason that GuestRoom couldn't be created (maybe they went in an invalid room id or something). Those bangs (!) can be super helpful in tests, OR they are helpful when you have exception handlers. If you're interested, take a look at exception handlers to see how you can utilize forced errors in your code.

app/models/guest.rb
```diff
@@ -0,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+end
```

app/models/guest_room.rb
```diff
@@ -0,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
```

app/models/room.rb
```diff
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
   belongs_to :hotel
+  has_many :guest_rooms
+  has_many :guests, through: :guest_rooms
 end
```
megstang, line 5: 👍

config/routes.rb
```diff
@@ -3,4 +3,6 @@
 
   # Defines the root path route ("/")
   # root "articles#index"
+  get "guests/:id", to: "guests#show"
+  post "guests/:id", to: "guest_rooms#create"
```
megstang, line 7: since this is going to the `guest_rooms` controller, which I agree with doing, it would make a bit more sense to use "guest_rooms" in your URL.

megstang: Or, you could do something like /guests/:id/rooms

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      it "I see a list of all the rooms they've stayed in including the room's suite, nightly rate, and the name of the hotel that it belongs to." do
+        visit "/guests/#{@guest_1.id}"
+        
+        expect(page).to have_content(@room_1.suite)
+        expect(page).to have_content(@room_1.rate)
+        expect(page).to have_content(@hotel_1.name)
+        expect(page).to_not have_content(@room_3.suite)
+        expect(page).to_not have_content(@room_3.rate)
+        expect(page).to_not have_content(@hotel_3.name)
+      end
```
megstang, line 38: Nice, I really like that you're testing what should also NOT be on the page.

megstang: I am wondering, though, if it would make sense to have more rooms for guest_1 so that we can make sure that the iterating on teh view works, and we can display more than just 1 room on the guests show page.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+          fill_in 'Add Room:', with: @room_2.id
+          click_button 'Submit'
+        end
+        save_and_open_page
+        expect(current_path).to eq("/guests/#{@guest_1.id}")
+        expect(page).to have_content(@room_2.suite)
+        expect(page).to have_content(@room_2.rate)
+        expect(page).to have_content(@hotel_1.name)
+      end
+    end
```
megstang, line 61: This first `it` block is not really necessary since you have the second one. The second `it` block will error out when it tries to fill in the field, and hitting the button if they are not present on the page. So, you could definitely get rid of that first one.

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,7 @@
+require "rails_helper"
+
+RSpec.describe Guest, type: :model do
+  describe "relationships" do
+    it { should have_many :rooms }
+  end
```
megstang, line 6: I know you're aware of this since we chatted a little after the assessment, but putting this here to keep feedback all in one spot-- make sure to test relationships for GuestRoom, Guests, and Rooms.

megstang (commented):
@fadwil I left some comments for you to take a look over. Let me know if you have any questions. I'll follow up in slack by EOD today with more feedback.

config/routes.rb
fadwil: do you mean like `post "guest_rooms/:id", to: "guest_rooms#create"` ? I guess what confuses me there is that I was working on the guest show page so I thought the route needed to be the same url.

If I did `/guests/:id/rooms` what else would I need to do, I know I would need to change the form in my show page to reflect that url, anything else?

spec/features/guests/show_spec.rb
fadwil: I was really hoping I could go back and add more to the tests if I had time, I had guest 2 in multiple rooms which I planned to test but didnt get far enough to go back and try.

config/routes.rb
megstang: The route on your form directs that form submission to an action that creates a new GuestRoom resource, and then that action redirects to the Guest show page.. so, they don't need to look the same. Your action will redirect where you need it to go.

If you did `/guests/:id/rooms`, you're right, you'd need to change the form and your routes file, but that should be it!

spec/features/guests/show_spec.rb
megstang: Totally get it... you were on limited time!

PR #23, opened 2023-07-24
show incomplete

Captainlearyo:
i got stuck at the end of story 1

### DIFF
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,0 +1,6 @@
+class GuestsController < ApplicationController
+  def show
+    @guests = Guest.find(params[:id]) 
+    
+  end
+end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,12 @@
+<h1><%= @guests.name %></h1>
+<p>Suite: <% @guests.rooms.each do |room| %></p>
+  <p><%= room.suite %></p>
+    <% end %>
+<p>Rate: <% @guests.rooms.each do |room| %></p>
+  <p><%= room.rate %></p>
+    <% end %>
+<p>Hotel name: <% @guests.rooms.each do |room| %></p>
+  <p><%= room.hotel_id %></p>
+    <% end %>
+
+
--- a/spec/features/guests/show_spec.rb
+++ b/spec/features/guests/show_spec.rb
@@ -1,0 +1,35 @@
+require "rails_helper"
+
+RSpec.describe "Guests show page" do
+  before(:each) do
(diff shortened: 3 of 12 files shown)


app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestsController < ApplicationController
+  def show
+    @guests = Guest.find(params[:id])
```
juliet-e, line 3: Just to not get confused, it might be helpful to name this `@guest` because it's just one single guest that's retrieve from the find method

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1><%= @guests.name %></h1>
+<p>Suite: <% @guests.rooms.each do |room| %></p>
+  <p><%= room.suite %></p>
+    <% end %>
+<p>Rate: <% @guests.rooms.each do |room| %></p>
+  <p><%= room.rate %></p>
+    <% end %>
+<p>Hotel name: <% @guests.rooms.each do |room| %></p>
+  <p><%= room.hotel_id %></p>
```
juliet-e, line 9: It looks like this part tripped you up on US 1. We are looking for the name of the hotel and it's actually easier than it seems -> because of the AR associations we've set up (`room` belongs to a hotel), we can call `room.hotel.name` and get the name of the hotel. Set up a pry in your controller or view and try calling this. It's easier than you're expecting!

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  #And I see a list of all the rooms they've stayed in
+  #including the room's suite, nightly rate, and the name of the hotel that it belongs to.
+    describe "When I visit a guest's show page" do
+      it "I see the guest's name, room's suite, nightly rate, name of hotel" do
+        visit "/guests/#{@guest_1.id}"
+        save_and_open_page
+        expect(page).to have_content(@guest_1.name)
+        expect(page).to have_content(@room_1.suite)
+        expect(page).to have_content(@room_1.rate)
+        expect(page).to have_content(@hotel_1.name)
```
juliet-e, line 23: You were off to a great start on this test! Nice work setting up lots of valuable test data

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "rails_helper"
+
+RSpec.describe Guest, type: :model do
+  describe "relationships" do
+    it {should have_many(:room_guests)}
```
juliet-e, line 5: Nice work with this model test. We'd also want to see a model spec for RoomGuest, and the additional relationships in the Room spec. It's a good idea to get these model specs set up before you even head into the first user story.

PR #24, opened 2023-07-24
Zanna Fitch

z-fitch:
I feel like it went pretty well. I finished the three required user stories and even tried the extension. I gotta practice sql and joins more, I kept getting tripped up and didn't know how to move forward. But I am proud that I attempted it :) I feel like I am a bit more comfortable with MVC conventions but still not confident when I make a decision on where to put certain things.

### DIFF
--- a/app/controllers/rooms_controller.rb
+++ b/app/controllers/rooms_controller.rb
@@ -1,0 +1,8 @@
+class RoomsController < ApplicationController
+  def index
+    @rooms = Room.all
+    # I was pretty sure that you could call the model method
+    #  in the veiw since its just reading the output 
+    #  and not dealing with data 
+  end
+end
--- a/app/models/hotel.rb
+++ b/app/models/hotel.rb
@@ -1,3 +1,11 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+  has_many :guests, through: :rooms
+
+  # def hotel_guests
+  #   select("guests.* FROM guests")
+  #   .joins("INNER JOIN guest_rooms ON guests.id = guest_rooms.guest_id")
+  #   .where("guest_rooms.room_id = hotel_id")
+  #   joins(:hotels).distinct.pluck(:name)
+  # end
 end
--- a/app/models/hotel.rb
+++ b/app/models/hotel.rb
@@ -1,6 +1,6 @@
 class Hotel < ApplicationRecord
   has_many :rooms
-  has_many :guests, through: :rooms
+
 
   # def hotel_guests
   #   select("guests.* FROM guests")
--- a/spec/features/guests/show_spec.rb
+++ b/spec/features/guests/show_spec.rb
@@ -1,0 +1,51 @@
+require "rails_helper"
+
+RSpec.describe "Show spec" do
+  before(:each) do
+    @emma = Hotel.create!(name: "Hotel Emma", location: "San Antonio")
+    @wolf = Hotel.create!(name: "Great Wolf Lodge", location: "Dallas")
+    
+    @room_1 = @emma.rooms.create!(rate: 500, suite: "Presidential")
+    @room_2 = @emma.rooms.create!(rate: 650, suite: "Penthouse")
+    @room_3 = @wolf.rooms.create!(rate: 300, suite: "Garden")
+    @room_4 = @wolf.rooms.create!(rate: 400, suite: "Animal Kingdom")
+    
+    @sara = Guest.create!(name: "Sara", nights: 2)
+    @leo = Guest.create!(name: "Leo", nights: 3)
+    @piper = Guest.create!(name: "Piper", nights: 3)
+    @paige = Guest.create!(name: "Paige", nights: 1)
+
+    @room_4.guests << @sara
+    @room_1.guests << @sara
+    @room_2.guests << @leo
+    @room_2.guests << @piper
+    @room_2.guests << @paige
+    @room_1.guests << @paige
+  end
+  # user story 1
+  describe "As a visitor, when going to the guest show page" do
+    it "has the guests name and a list of all rooms they are in:shows room's suite, nightly rate, and the name of the hotel that it belongs to" do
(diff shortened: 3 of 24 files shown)


app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,8 @@
+class RoomsController < ApplicationController
+  def index
+    @rooms = Room.all
+    # I was pretty sure that you could call the model method
```
juliet-e, line 4: You're correct! Referencing the attributes and the model method you created is totally fine and encouraged

app/models/hotel.rb
```diff
@@ -1,3 +1,11 @@
 class Hotel < ApplicationRecord
   has_many :rooms
+
+
+  # def hotel_guests
```
juliet-e, line 5: Something I like to think about when doing complex joins is starting LEFT and going RIGHT, where left is the model I'm starting at (Hotel) and Right is the table I'm trying to get to (guests). 
In this case, we can join hotels to rooms because of the one-to-many relationship, and then using the joins table under the hood, we can join rooms to guests. In ActiveRecord, we don't even need to include an explicit join with the joins table as AR will do that implicitly because you've set up relationships to allow guests to associate to rooms (`has_many :rooms, through :guest_rooms`). 
The AR query you could do here that works is:
`Hotel.joins(rooms: :guests).distinct("guests.name").pluck("guests.name")`

When I do that in rails console, the SQL given is:
`SELECT DISTINCT "guests"."name" FROM "hotels" INNER JOIN "rooms" ON "rooms"."hotel_id" = "hotels"."id" INNER JOIN "guest_rooms" ON "guest_rooms"."room_id" = "rooms"."id" INNER JOIN "guests" ON "guests"."id" = "guest_rooms"."guest_id"`
It looks complex, but it's just doing a chain of joins from left to right: hotels -> rooms -> guest_rooms -> guests

Another option that is way simpler is a little ActiveRecord magic. Because you have the one-to-many relationship between hotels and rooms, you could add this to the Hotel model: `has_many :guests, through: :rooms`
This will allow you to do `hotel.guests.distinct` and achieve the same result. We haven't talked about this much, but it's possible to shortcut the joining by adding this association, which associates another model to a given model if that other model is associated to the given model's child.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    @room_1.guests << @sara
+    @room_2.guests << @paige
+    @room_1.guests << @paige
+  end
+  # user story 1
+  describe "As a visitor, when going to the guest show page" do
+    it "has the guests name and a list of all rooms they are in:shows room's suite, nightly rate, and the name of the hotel that it belongs to" do
+      visit "/guests/#{@paige.id}"
+
+      expect(page).to have_content(@paige.name)
```
juliet-e, line 28: These assertions could be more robust if you verify the text is there as it is displayed in your view, so:
`expect(page).to have_content("Hotel Emma, Penthouse, Room Rate: 650")`
By doing this, you're making sure the dynamic parts of this page are populating properly and *in the right place* rather than floating randomly on the page

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    it " show the number of guests that stayed in each room" do 
+      visit "/rooms"
+
+      expect(page).to have_content(@room_1.suite)
+      expect(page).to have_content(@room_1.rate)
+      expect(page).to have_content(@room_1.hotel.name)
+      # Idk how you guys want that tested here so heres 3 ways
+      expect(page).to have_content(@room_1.guests.count)
```
juliet-e, line 49: The assertion on line 50 is the best because it makes sure that the guest count shows up next to the `Guest Count: ` text, rather than there being a floating 2 randomly on the page. 

It's also best to test in this way rather than verifying that it equals `@room_1.guests.count` because the `guests.count` is what you have in your implementation in the model method. When testing, we don't want to use the same code that's used in the implementation because the calculations in both places will always be the same, but not necessarily indicate whether the functionality is correct or expected. For example, given this method
`def add(a, b)`
`   a + b`
`end`

When we're testing this, we would want to test it like this: `expect(add(1,2)).to eq(3)` instead of `expect(add(1,2)).to eq(1 + 2)`. In the second case, we're just using the same calculation that's in the code. If it's *wrong* and not the right implementation, our test wouldn't catch it because it's just the same calculation in the test. 

Let me know if you have follow up questions!

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    @room_1.guests << @sara
+    @room_2.guests << @paige
+    @room_1.guests << @paige
+  end
+#user story 3
+  describe " When I visit the rooms index page" do
+    it "has a list of all rooms and its attributes and its hotel" do
+      visit "/rooms"
+
+      expect(page).to have_content(@room_1.suite)
```
juliet-e, line 28: With this test, we can't really tell if any of the room names, rates or hotels are grouped in the right way. This is a good area to use within blocks to make the test more meaningful!

spec/models/room_spec.rb
```diff
@@ -3,5 +3,38 @@
...
+    @room_4.guests << @sara
+  end
+
+  describe "Instance methods" do
+    describe "#guest_count" do
+      it "counts the number of guests in a room" do 
+        expect(@room_1.guest_count).to eq(2)
+        expect(@room_2.guest_count).to eq(1)
+        expect(@room_3.guest_count).to eq(1)
+        expect(@room_4.guest_count).to eq(1)
```
juliet-e, line 36: Great work testing these model methods!

app/models/hotel.rb
juliet-e: oops my query wasn't totally correct here! should be 
`Hotel.joins(rooms: :guests).where("rooms.hotel_id = #{hotel.id}").distinct("guests.name").pl
uck("guests.name")` so that you get the specific hotel id

PR #25, opened 2023-07-24
Kaina Cockett

kcockett:
I struggled way too long (90 minutes) with the model setups.  Once I had those down, I got through the rest of the user stories with very little problem.  I'll have to study the model associations better.


db/schema.rb
```diff
@@ -10,10 +10,33 @@
...
+
+  create_table "hotel_rooms", force: :cascade do |t|
+    t.bigint "hotel_id", null: false
+    t.bigint "room_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["hotel_id"], name: "index_hotel_rooms_on_hotel_id"
+    t.index ["room_id"], name: "index_hotel_rooms_on_room_id"
+  end
+
```
megstang, line 39: You do not need this table. I see you got your model set up for it, but only `GuestRoom` was the necessary join table.

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    redirect_to "/guests/#{@guest.id}"
+  end
+  def add_room
+    @guest = Guest.find(params[:guest_id])
+    room = Room.find(params[:room_id])
+    @guest.rooms << room
+
+    redirect_to "/guests/#{@guest.id}"
+  end
```
megstang, line 18: The only method that will probably function how you want it to is the `add_room` method. That said, what are we creating in this method? We're creating a GuestRoom object. For that reason, it would be most ReSTful to have this in a `GuestRoomsController` in a `create` action. 

Also, we could make this code a bit more optimal. Right now, you're finding the guest and finding the room, and then creating the object. But, you actually have EVERYTHING you need to make that connection without first finding those two objects. You could do the following: 

```ruby
GuestRoom.create(guest_id: params[:guest_id], room_id: params[:room_id])
redirect_to "/guests/#{params[:guest_id]}"
```

megstang: Notice in that snippet, I'm not storing anything into any instance variables since I don't care for any variables to get sent to a view, since I'm redirecting to a different page.

app/models/guest.rb
```diff
@@ -0,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+end
```

app/models/guest_room.rb
```diff
@@ -0,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
```

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,15 @@
...
+    <%= room.hotel.name %>
+  <% end %>
+</section>
+
+<section id="add_room_form">
+  <%= form_with url: add_room_to_guest_path(@guest), method: :post, local: true do |form| %>
+    <%= form.text_field :room_id, value: "Enter Room ID" %>
+    <%= form.submit "Add Room" %>
+  <% end %>
+</section>
```
megstang, line 15: 👍 Form looks good! Nice use of a path helper.

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<% @rooms.each do |room| %>
+  <%= room.suite %>
+  <%= room.rate %>
+  <%= room.hotel.name %>
+  
+<% end %>
```
megstang, line 6: We also want to see on this page, the number of guests that have stayed in each room.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+        expect(page).to have_content(@guest_1.name)
+      end
+
+      it "And I see a list of all the rooms they've stayed in including the room's suite, nightly rate, and the name of the hotel that it belongs to." do
+        visit "/guests/#{@guest_1.id}"
+
+        expect(page).to have_content(@room_1.suite)
+        expect(page).to have_content(@room_1.rate)
+        expect(page).to have_content(@room_1.hotel.name)
+      end
```
megstang, line 25: since you are testing that ALL of their rooms appear on the page, it would be great to have guest_1 have multiple rooms so could could test that all of their rooms are appearing, and not just one.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      it "When I fill in a field with the id of an existing room And I click submit Then I am redirected back to the guest's show page And I see the room now listed under this guest's rooms." do
+        visit "/guests/#{@guest_1.id}"
+        fill_in :room_id, with: "#{@room_1.id}"
+        click_button "Add Room"
+
+        expect(current_path).to eq("/guests/#{@guest_1.id}")
+        within("#room_list") do
+          expect(page).to have_content("#{@room_1.suite}")
+        end
+      end
```
megstang, line 42: In your setup, you are already adding room_1 to guest_1, so this test does not exactly test this functionality since room1 is already associated with guest1 before this test is run.

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    @hotel_1 = Hotel.create!(name: "Echo Mountain Inn", location: "Echo Mountain")
+    @room_1 = Room.create!(rate: 125, suite: "Presidential", hotel: @hotel_1)
+    @guest_1 = Guest.create!(name: "Charlize Theron", nights: 3)
+    @guest_1_room_1 = GuestRoom.create!(guest_id: @guest_1.id, room_id: @room_1.id)
+  end
+  describe "As a visitor" do
+    describe "When I visit the rooms index page" do
+      it "Then I see a list of all rooms including the room's suite, nightly rate, and the name of the hotel that it belongs to and the number of guests that have stayed in that room." do
+        visit "/rooms"
+        save_and_open_page
```
megstang, line 14: when you're pushing up code, take these methods out so it doesn't affect when other devs pull down your code and run your test suite.

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    describe "When I visit the rooms index page" do
+      it "Then I see a list of all rooms including the room's suite, nightly rate, and the name of the hotel that it belongs to and the number of guests that have stayed in that room." do
+        visit "/rooms"
+        save_and_open_page 
+        expect(page).to have_content("#{@room_1.suite}")
+        expect(page).to have_content("#{@room_1.rate}")
+        expect(page).to have_content("#{@room_1.hotel.name}")
+        
+      end
+    end
```
megstang, line 20: A similar comment to before, if you're creating funtionality such that you have an index of something, make sure you have more than one of that resource so you can make sure that you truly are displaying ALL of that resource.

megstang (commented):
@kcockett I left some comments for you to look over. Please let me know if you have any questions! I'll go ahead and get some more feedback to you before EOD today via slack.

PR #26, opened 2023-07-24
Gabe

Gabe-Torres:
Something happened during my migrations and table creations. Well, I created them incorrectly. I didn't get that fixed until it was to late and it put me behind during the IC.


app/models/hotel.rb
```diff
@@ -1,3 +1,4 @@
 class Hotel < ApplicationRecord
-  has_many :rooms
+  has_many :rooms, dependent: :destroy 
+  has_many :guests, through: :rooms
```
cjsim89, line 3: Nice job creating this association! It's a really nice shortcut that I use all the time in my model methods.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,11 @@
+<h1><%= @guest.name %></h1>
+
+<section class="guests">
+  <% @guest.guest_rooms.each do |guest_room| %>
```
cjsim89, line 4: This makes sense, but your Guest model also `has_many :rooms` so instead of going to the join table, you can go directly to the `rooms` association and save space in your view. Your approach isn't wrong, just can be more efficient.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,11 @@
...
+    <section id="guest-room-<%= guest_room.id %>">
```
cjsim89, line 5: Also, be careful with this  - this `guest_room.id` is the primary key for the join table, not a guest or room.

db/migrate/20230724171501_create_guests.rb
```diff
@@ -0,0 +1,11 @@
+class CreateGuests < ActiveRecord::Migration[7.0]
+  def change
+    create_table :guests do |t|
+      t.string :name
+      t.integer :nights
+      t.references :hotel, null: false, foreign_key: true
```
cjsim89, line 6: This would be fine if there were a one-to-many relationship between a Guest and a Hotel, but like this it means that a Guest can only ever belong to one hotel. Poor Charlize Theron can only ever stay at the Hilton! 😢 

Instead, we have a many-to-many relationship through the join table, and through that relationship we can figure out which guest stayed at which hotel (through the rooms).

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+# I see the guest's name
+# And I see a list of all the rooms they've stayed in
+# including the room's suite, nightly rate, and the name of the hotel that it belongs to.
+  describe "as a visitor" do
+    it "all guest info" do 
+      visit "/guests/#{@guest_1.id}"
+
+      expect(page).to have_content(@guest_1.name)
+
+      within("#guest-room-#{@guest_room.id}") do
```
cjsim89, line 25: Good job with this within block! Keep this up

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+RSpec.describe "Guests show page", type: :feature do 
+  before(:each) do
+    @hotel_1 = Hotel.create!(name: "Tipton Hotel", location: "Boston, Massachusetts")
+    @room_1 = @hotel_1.rooms.create!(rate: 300, suite: "standard")
+    @guest_1 = @room_1.guests.create!(name: "Zack", nights: 4)
+    GuestRoom.create!(room_id: @room_1.id, guest_id: @guest_1.id)
```
cjsim89, line 8: Good object setup - might be useful to have more than 1 instance of each object, just to have more stuff to play with.

spec/models/room_spec.rb
```diff
@@ -3,5 +3,6 @@
 RSpec.describe Room, type: :model do
   describe "relationships" do
     it { should belong_to :hotel }
+    it { should have_many :guests }
```
cjsim89, line 6: There's one more association/relationship you're missing; `should have_many :guest_rooms` so that your line 6 looks like `it { should have_many(:guests).through(:guest_rooms) } `.

PR #27, opened 2023-07-24
Bradley Milton

brad2412:
Good morning!  I was only able to make it through user story 2.  I had a Physical Therapy appointment that I thought I would be able to make it back at the start of class.  I didn't get my full 3 hours to complete but think I would have completed the whole thing if time was better.  I did run into a slow start with setting up associations but was able to start rocking when I set it up right.  I look forward to all the feedback on what I could improve upon next time!  Have a great day!

### DIFF
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,0 +1,6 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms_stayed_in = @guest.rooms
+  end
+end
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,6 +1,16 @@
 class GuestsController < ApplicationController
+
   def show
     @guest = Guest.find(params[:id])
     @rooms_stayed_in = @guest.rooms
+  end
+
+  def create
+    @guest = Guest.find(params[:id])
+    room = Room.find(params[:guest][:room_id])
+    
+    @guest.rooms << room
+
+    redirect_to "/guests/#{@guest.id}"
   end
 end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,13 @@
+<h1><%= @guest.name %></h1>
+<h2>Rooms stayed in:</h2>
+<ul>
+<section id="rooms">
+  <% @rooms_stayed_in.each do |room| %>
+    <li>
+      Suite: <%= room.suite %>
+      Rate: <%= room.rate %>
+      Hotel: <%= room.hotel.name %>
+    </li>
+  <% end %>
+</ul>
+</section>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -11,3 +11,11 @@
   <% end %>
 </ul>
 </section>
+
+<section id="add-room-form">
+  <%= form_with  url: "/guests/#{@guest.id}", local: true, method: :post do |form| %>
+    <%= form.label :room_id, "Room ID" %>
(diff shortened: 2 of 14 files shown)


app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  def create
+    @guest = Guest.find(params[:id])
+    room = Room.find(params[:guest][:room_id])
+    
+    @guest.rooms << room
+
+    redirect_to "/guests/#{@guest.id}"
+  end
```
megstang, line 15: Since you're creating a GuestRoom resource, this should be in a GuestRoomsController. 

Also, we could make this code a bit more optimal by creating the resource right away, rather than finding two objects and making an association. It could look something like this: 

```ruby
    GuestRoom.create(guest_id: params[:id], room_id: params[:room_id])
```

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,16 @@
+class GuestsController < ApplicationController
+
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms_stayed_in = @guest.rooms
+  end
```
megstang, line 6: You can eliminate line 5 since you can get rooms off of the guest object that you're already passing to the view. We want to send as few objects to the view as possible.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,21 @@
+<h1><%= @guest.name %></h1>
+<h2>Rooms stayed in:</h2>
+<ul>
+<section id="rooms">
+  <% @rooms_stayed_in.each do |room| %>
```
megstang, line 5: So here, following up on my earlier comment on sending only the guest object to the view, we'd just call `@guest.rooms.each ...`

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,21 @@
...
+</ul>
+</section>
+
+<section id="add-room-form">
+  <%= form_with  url: "/guests/#{@guest.id}", local: true, method: :post do |form| %>
+    <%= form.label :room_id, "Room ID" %>
+    <%= form.text_field :room_id, name: "guest[room_id]"%>
+    <%= form.submit "Add Room" %>
+  <% end %>
+</section>
```
megstang, line 21: Your route on line 16 could be a bit more restful since you're actually creating a GuestRoom. It could go to something like '/guest_rooms' and pass the guest id via a hidden field in the form. Or, it might even make more sense to have something like "/guests/:id/rooms"

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    expect(current_path).to eq "/guests/#{guest.id}"
+
+    within("#rooms") do
+      expect(page).to have_content("Suite: Deluxe Suite")
+      expect(page).to have_content("Rate: 150")
+      expect(page).to have_content("Hotel: Continental")
+    end
+  end
+end
```
megstang, line 40: Nice use of `within`'s!

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    expect(page).to have_content("John Wick")
+    expect(page).to have_content("Rooms stayed in:")
+
+    within("#rooms") do
+      expect(page).to have_content("Suite: President Suite")
+      expect(page).to have_content("Rate: 100")
+      expect(page).to have_content("Hotel: Continental")
+    end
+  end
```
megstang, line 20: It would be good to get some more rooms for the guest so you can make sure ALL of their rooms can get displayed, and that it doesnt just display one room.

megstang (commented):
@brad2412 Let me know if you have any questions on any of my comments! I'll follow up in slack shortly.

